PR #9, opened 2020-07-23
Refactoring - Part II

frapac:
This PR adds more refactoring to ExaPF. In short: 
- we rename `TimerOutputs`'s `to` to `TIMER`, as we had clashes with local variables named `to` previously
- we improve the handling of submodules in the package, to ensure that all modules are using the same types 
- we move all iterative solvers and preconditioning into the submodule `Iterative`, by adding a dedicated function `ldiv!`. Multiple dispatch could be use further here, but that will be the goal of a future PR

### DIFF
--- a/src/ExaPF.jl
+++ b/src/ExaPF.jl
@@ -356,6 +356,7 @@ function solve(pf::Pf, npartitions=2, solver="default";
 
     J = residualJacobian(V, Ybus, pv, pq)
     dim_J = size(J, 1)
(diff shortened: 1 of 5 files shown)


dmaldona:
@frapac, @michel2323 I added myself as reviewer just to make sure I read the commits and don't lose track.

frapac:
@dmaldona I thought that I flag you as a reviewer when opening this PR, but it looks like my suggestion was not registered by Github... Sorry about that!

PR #11, opened 2020-07-24
Refactor solver (Part I): Moved power-system data structures out of solve function

dmaldona:
Hi guys,

This is my first stab refactoring the solve function. Here I just have created a new submodule (PowerSystems) with the Pf data structure and a constructor that creates the impedance matrix, voltage, etc...

Since there might be many breaking changes I am doing this refactoring in small parts. Let me know hat you think.


Project.toml
```diff
@@ -13,6 +13,7 @@ LineSearches = "d3d80556-e9d4-5f37-9878-2ab0fcc64255"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Metis = "2679e427-3c69-5b7f-982b-ece356f1e94b"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
+Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
```
frapac, line 16: I would maybe not include `Revise` in `Project.toml`

src/ExaPF.jl
```diff
@@ -36,15 +36,11 @@ include("iterative.jl")
 using .Iterative
 include("network.jl")
 using .Network
-
+include("psystem.jl")
```
frapac, line 39: Do you know if we could rename `psystem.jl` `powersystem.jl` (to have more explicit namings)?

src/psystem.jl
```diff
@@ -0,0 +1,153 @@
...
+"""
+    Pf
+
+This structure contains constant parameters that define the topology and
+physics of the power network.
+
+# Fields
+- `V::Array{Complex{Float64}}`: voltage
+"""
+struct Pf
```
frapac, line 17: How about renaming `Pf` `PowerFlow`?

src/psystem.jl
```diff
@@ -0,0 +1,153 @@
...
+using ..ExaPF: Parse
+using ..ExaPF: Network
+
+using SparseArrays
+
+"""
+    Pf
+
+This structure contains constant parameters that define the topology and
+physics of the power network.
```
frapac, line 12: As far as I understand, `Pf` creates object in the main memory, and not directly onto the GPU. I would maybe add a comment about that here, just to tell the user that some objects are bound to be exported to the GPU after.

src/psystem.jl
```diff
@@ -0,0 +1,153 @@
...
+    
+        Sbus = assembleSbus(gen, load, SBASE, nbus)
+        
+        new(V, Ybus, data, nbus, ngen, nload, ref, pv, pq, Sbus)
+    end
+end
+
+
+"""
+bustypeindex(data)
```
frapac, line 73: Maybe indent this line for formatting?

src/psystem.jl
```diff
@@ -0,0 +1,153 @@
...
+    ref = findall(x->x==3, bustype)
+    pv = findall(x->x==2, bustype)
+    pq = findall(x->x==1, bustype)
+
+    return ref, pv, pq
+end
+
+
+"""
+assembleSbus(data)
```
frapac, line 119: I would maybe add an indent as well. Also, it looks like the signature does not match those of the mentioned function.

michel2323:
Maybe we should also put `Parse` and `Network` into the powersystem module.

src/ExaPF.jl
dmaldona: Ahh, oversight.

src/psystem.jl
dmaldona: Renamed it to PowerNetwork (referring to the physical object)

src/ExaPF.jl
dmaldona: Done

Project.toml
dmaldona: Oversight! It happens too often. Fixed it!

src/psystem.jl
dmaldona: Done

src/psystem.jl
dmaldona: Done

src/psystem.jl
dmaldona: Added comment

dmaldona:
@michel2323 cannot respond directly to your comment. 

Good idea. I have moved `Network` into the powersystem module. The other module, `Parse`, its a bit more trickier to I will create an issue and come back to it later.

dmaldona:
@michel2323 , @frapac : I have pushed a new commit addressing your comments. Please let me know if OK,

frapac:
Looks good to me!

PR #14, opened 2020-07-29
Add more unit-tests

frapac:
This PR adds more unit-tests to ExaPF
- we aim at adding "bottom-up" testing before testing the `solve` function by itself. The idea is to fail fast if a problem occurs (if `mul!` is broken in CUDA.jl, I want to detect it before testing the function `solve`)
- we push also small fixes in the core of ExaPF

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,24 +1,53 @@
+using LinearAlgebra
+using Random
+using SparseArrays
 using Test
+using TimerOutputs
 using CUDA
 
 # This is a problem of the code right now. It can only set once per as
 # this variable is used in macros to generate the code at compile time.
 # This implies we cannot both test gpu and cpu code here.
 target = "cpu"
+using ExaPF
+
+@testset "Iterative solvers" begin
+    n, m = 32, 32
+    # Add a diagonal term for conditionning
+    A = randn(n, m) + 15I
+    x♯ = randn(m)
+    b = A * x♯
+    # Careful: all algorithms work with sparse matrix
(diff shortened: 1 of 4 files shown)


test/runtests.jl
```diff
@@ -1,25 +1,124 @@
...
+            jacobianAD, ExaPF.residualFunction_polar!, Vm, Va,
+            ybus_re, ybus_im, pbus, qbus, pv, pq, nbus, to)
+        @test jacobianAD.J ≈ J♯
+    end
+end
+
+@testset "Wrapping of iterative solvers" begin
+    n, m = 32, 32
+    # Add a diagonal term for conditionning
+    A = randn(n, m) + 15I
```
dmaldona, line 86: @frapac can you seed Random by default for reproducibility?

frapac: Good idea :+1:

michel2323:
Does someone know whether there is an easy way to add optional tests? So the 70k case could be run by a user. It would also be good to have a timing test, such that we don't lose performance over time.

frapac:
Good idea for the timing test. An idea would be to do something similar as MOI's benchmarks, as implemented [here](https://github.com/jump-dev/MathOptInterface.jl/blob/master/perf/bellman_ford.jl). For the optional tests, I am not aware of any proper solution though...

michel2323:
Added Krylov.dqgmres, set a restart=4 for all gmres's as otherwise they don't converge with bigger cases. I also ran all this on Summit.

PR #16, opened 2020-07-31
Refactor solve (Part 2): creating vectors x, u, p

dmaldona:
This is a rough pull request to obtain your comments and suggestions.

In this PR I have created two functions that act as map and inverse map between the physical variables that pertain to each bus:

- voltage magnitude (vmag)
- voltage angle (vang)
- active power injection (pinj)
- reactive power injection (qinj)

and the mathematical variables (x, u, p) that are used by the reduced gradient method. We want to be able to write:

g(x, u, p) = 0

and then obtain dg/dx and dg/du. Our original code was based on Matpower's implementation. Matpower uses the physical variables since this allows writing some expressions in a very succinct way. For instance, if we use complex voltage (Vbus = vmag + 1j*vang) and complex power injection (Sbus = P + j*Q) then g(x, u, p) can be written:

V .* conj(Ybus * V) - Sbus

The question now is how to refactor solve(). I think the kernels are easier to write using the physical variables (vmag, vang, pinj, qinj) as these are indexed by bus number. Writing g() using x, u, p directly requires a somewhat involved indexing that might not very GPU friendly.


michel2323:
In my experience views work pretty well on GPUs:

```julia
vals = [2.0, 5.0, 3.0 , 4.0]
xidx = [1,3,5]
uidx = [2,4,5]
x = view(vals, xidx)
u = view(vals, uidx)
```
If all three arrays are on the GPU, that is pretty fast and works with AD and everything. One may lose some performance, but in `ACOPF.jl` AD was still flying. The nice bit is that when you change `vals`, `x` and `u` change too.

src/powersystem.jl
```diff
@@ -60,15 +62,194 @@ struct PowerNetwork
...
+end
+
+"""
+    view(PowerNetwork)
+
+Prints power network characteristics.
 
-        new(V, Ybus, data, nbus, ngen, nload, ref, pv, pq, Sbus)
+"""
+
```
frapac, line 81: I would remove the blank line between the docstring and the function

src/powersystem.jl
```diff
@@ -60,15 +62,194 @@ struct PowerNetwork
...
+function view(pf::PowerNetwork)
```
frapac, line 82: That's nice to have this function! I would be in favor of overloading `Base.show` so we could just do `println(pf)`. 
```
import Base: show
function Base.show(io::IO, pf::PowerNetwork)
     ... the code you implemented here
end
```

src/powersystem.jl
```diff
@@ -60,15 +62,194 @@ struct PowerNetwork
...
+                type, vmag, vang, pinj, pinj)
     end
+
 end
 
 
+
+
+"""
+    assemble_vecs(PowerNetwork)
```
frapac, line 110: I think we should think about a more explicit naming. How about `partition`? Or split in three functions `state`, `control`, `variable`? Or just overloading a function `get` so that we just have to type `get(pf::PowerNetwork, State())`?

src/powersystem.jl
```diff
@@ -60,15 +62,194 @@ struct PowerNetwork
...
+        end
+    end
+
+    return x, u, p
+
+end
+
+
+"""
+    xup_to_vs(PowerNetwork, x, u, p)
```
frapac, line 199: I think we should also think about the naming for this function :)

src/powersystem.jl
```diff
@@ -60,15 +62,194 @@ struct PowerNetwork
...
+        end
+    end
+
+    # build vector u
+    dimension = 2*npv + nref
+    u = zeros(dimension, 1)
+    k = 1
+
+    for bus=1:pf.nbus
+        if pf.bustype[bus] == 2 #PV
```
frapac, line 167: I would be in favor of specifying the PV, PQ and SLACK types as global constant, to clarify things a bit. 
e.g. by adding at the beginning of this file:
```julia
const PQ_BUS_TYPE = 1
const PV_BUS_TYPE = 2
const SLACK_BUS_TYPE  = 3
```

src/powersystem.jl
```diff
@@ -60,15 +62,194 @@ struct PowerNetwork
...
+            k = k + 1
+        elseif pf.bustype[bus] == 1 # PQ
+            p[k] = real(pf.Sbus[bus])
+            p[k + 1] = imag(pf.Sbus[bus])
+            k = k + 2
+        end
+    end
+
+    return x, u, p
+
```
frapac, line 194: Maybe remove this blank line?

src/powersystem.jl
```diff
@@ -60,15 +62,194 @@ struct PowerNetwork
...
+            vang[bus] = p[p_idx]
+            u_idx += 1
+            p_idx += 1
+        end
+
+    end
+
+
+    return vmag, vang, pinj, qinj
+
```
frapac, line 245: Maybe remove these two blank lines?

src/powersystem.jl
```diff
@@ -60,15 +62,194 @@ struct PowerNetwork
...
+end
+
+
```
frapac, line 248: I would be in favor of having only one blank line between two functions

src/powersystem.jl
```diff
@@ -60,15 +62,194 @@ struct PowerNetwork
...
+Vector p is the parameter vector, consisting on:
+    - VA for buses of type SLACK (type 3)
+    - P, Q for buses of type PQ (type 1)
+These parameters are fixed through the problem.
+
+In addition, variables:
+    - Q for buses of type PV (type 2)
+    - P, Q for buses of type SLACK (type 3)
+are explicitly obtained from the previous variables, and hence
+not needed in the computation.
```
frapac, line 135: Nice docstring!

src/powersystem.jl
dmaldona: @frapac I agree. Is a terrible name. It was late at night...

src/powersystem.jl
dmaldona: @frapac Ill do the get overloading as you proposed!

src/powersystem.jl
dmaldona: @frapac yes, you're right. I'll change that.

src/powersystem.jl
dmaldona: @frapac sure! can we use an enum? https://docs.julialang.org/en/v1/base/base/#Base.Enums.Enum

src/powersystem.jl
dmaldona: @frapac sorry Ill pay more attention to the style guide in the future.

src/powersystem.jl
dmaldona: @frapac Thanks!

src/powersystem.jl
frapac: To be fairly honest, I am bad at naming even during the day ^^

src/powersystem.jl
frapac: Indeed, an enum would be more appropriate!

dmaldona: Ended up declaring as global const

dmaldona:
@michel2323 @frapac : addressed review comments. I did "get_u", "get_v" and "get_p" instead of multiple dispatch.

Project.toml
```diff
@@ -15,6 +15,7 @@ LineSearches = "d3d80556-e9d4-5f37-9878-2ab0fcc64255"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Metis = "2679e427-3c69-5b7f-982b-ece356f1e94b"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
+Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
```
frapac, line 18: I would remove `Revise` here

examples/pf.jl
```diff
@@ -12,5 +12,8 @@ function pf(datafile, npartition=2, solver="default")
     println(solver)
     
     pf = PowerSystem.PowerNetwork(datafile)
-    return solve(pf, npartition, solver);
+    x = ExaPF.PowerSystem.get_x(pf)
+    u = ExaPF.PowerSystem.get_u(pf)
+    p = ExaPF.PowerSystem.get_p(pf)
```
frapac, line 17: Nice! Why we do not use multiple dispatch at the end? (I think both solutions are valid, though)

src/ExaPF.jl
```diff
@@ -188,7 +188,7 @@ function polar!(Vm, Va, V, ::CUDADevice)
     Va .= CUDA.angle.(V)
 end
 
-function solve(pf::PowerSystem.PowerNetwork, npartitions=2, solver="default";
+function solve(pf::PowerSystem.PowerNetwork, x, u, p, npartitions=2, solver="default";
```
frapac, line 191: I would maybe add proper type annotation for vector there: 
```julia
x::AbstractArray,
u::AbstractArray,
p::AbstractArray,
```

Project.toml
dmaldona: I keep doing this :(

examples/pf.jl
dmaldona: I had some questions about it but since I did it on Saturday I didnt want to bother anyone. Can we talk about it in our next meeting? Ill push it like this for now and later I can change it to the multiple dispatch version.

src/ExaPF.jl
dmaldona: Right!

PR #18, opened 2020-08-05
New u, x, p indexing

dmaldona:
As we discussed, this new indexing is in accord with the document we are preparing.

### DIFF
--- a/src/powersystem.jl
+++ b/src/powersystem.jl
@@ -100,7 +100,7 @@ function Base.show(io::IO, pf::PowerNetwork)
         pinj = real(pf.Sbus[i])
         qinj = imag(pf.Sbus[i])
         @printf("\t%i \t  %d \t %1.3f\t%3.2f\t%3.3f\t%3.3f\n", i,
-                type, vmag, vang, pinj, pinj)
+                type, vmag, vang, pinj, qinj)
     end
 
 end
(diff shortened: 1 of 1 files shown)


src/powersystem.jl
```diff
@@ -151,8 +147,11 @@ Vector u is the control vector, consisting on:
     - VMAG, P for buses of type PV (type 1)
     - VM for buses of type SLACK (type 3)
 These variables are controlled by the grid operator.
-"""
 
+Ordering:
+
+u = [VMAG^{REF}, P^{PV}, V^{PV}]
```
frapac, line 153: I have a naive question here: could we put `VMAG^{REF}` to 0, as it is the reference voltage?

dmaldona: @frapac Good question. Every voltage limit is enforced to be between 0.9 and 1.1 (because of physical considerations). Furthermore, the power flow equations are ill conditioned when vmag approaches 0 (low voltage conditions). In general, we often set VMAG ^{REF} to be 1.0.

frapac: If VMAG^{REF} is constant, does it make sense to include it in `u`?

dmaldona: VMAG^{REF} can be adjusted/controlled. Sorry, before I referred to the power flow problem (solution of the power flow equations). In OPF we do change the value of VMAG^{REF}. However, VANG^{REF} is constant both in the PF and the OPF problem.

PR #19, opened 2020-08-05
Added design variable Jacobian

michel2323:
This implements the design Jacobian. However, it's not beautiful. I had to essentially duplicate the AD code because of the different indexing of the variables. I guess there's a smarter way to do this. Let's verify whether this is anything useful.

The design Jacobian is dense for now, since I don't know the sparsity pattern. So it won't work for large cases, due to lack of speed and memory.

This is not yet ready to be merged to master.

### DIFF
--- a/examples/pf.jl
+++ b/examples/pf.jl
@@ -1,13 +1,14 @@
 using ExaPF
 using ExaPF.Parse
 using ExaPF.PowerSystem
+using KernelAbstractions
 
 # file locations
 # datafile = "GO-Data/datasets/Trial_3_Real-Time/Network_30R-025/scenario_1/case.raw"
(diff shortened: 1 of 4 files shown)


michel2323:
I think I can put all the code of both Jacobians back into one. One would then have to pass the mapping to the Jacobian. The 2 Jacobian computations only differ in the mapping:

```julia
mappv =  [i + nv_a for i in pv]
map = T{Int64}(vcat(ref, mappv, pv))
```
and
````julia
mappv = [i + nv_m for i in pv]
mappq = [i + nv_m for i in pq]
map = T{Int64}(vcat(mappv, mappq, pq))
```

PR #23, opened 2020-08-12
Refactor [Part 3]: add matpower parse capabilities. Sketch cost function

dmaldona:
This pull request is still a work in progress.

I have added support for reading Matpower files. The user now can select to read either a PSSE or Matpower case when creating the power system object.

Before merging, I will complete the cost function.


src/ExaPF.jl
```diff
@@ -24,14 +24,18 @@ using TimerOutputs
...
 
 # Import submodules
-include("parse/parse.jl")
-using .Parse
 include("ad.jl")
 using .AD
 include("algorithms/precondition.jl")
 using .Precondition
+include("indexes.jl")
+using .IdxSet
```
frapac, line 32: I would be in favor of choosing a more explicit name for this module. How about `IndexSet`? `IndexingSet`?

src/indexes.jl
```diff
@@ -0,0 +1,102 @@
...
+    ZONE        = 11   # zone, loss zone (1-999)
+    VMAX        = 12   # maxVm, maximum voltage magnitude (p.u.)      (not in PTI format)
+    VMIN        = 13   # minVm, minimum voltage magnitude (p.u.)      (not in PTI format)
+
+    LAM_P       = 14   # Lagrange multiplier on real power mismatch (u/MW)
+    LAM_Q       = 15   # Lagrange multiplier on reactive power mismatch (u/MVAr)
+    MU_VMAX     = 16   # Kuhn-Tucker multiplier on upper voltage limit (u/p.u.)
+    MU_VMIN     = 17   # Kuhn-Tucker multiplier on lower voltage limit (u/p.u.)
+
+    return BUS_I, BUS_TYPE, PD, QD, GS, BS, BUS_AREA, VM, VA, BASE_KV, ZONE, VMAX, VMIN,
```
frapac, line 25: This function returns a lot of indexes. How about returning a single named tuple, to ease the retrieval? See e.g. https://docs.julialang.org/en/v1/manual/types/#Named-Tuple-Types-1

src/parse/parse_mat.jl
```diff
@@ -0,0 +1,1667 @@
...
+
+using ..ExaPF: IdxSet
+
+include("matlab.jl")
+include("im_common.jl")
+
+"Parses the matpwer data from either a filename or an IO object"
+function parse_matpower(io::IO)::Dict
+    mp_data = _parse_matpower_string(read(io, String))
+    #pm_data = _matpower_to_powermodels!(mp_data)
```
frapac, line 20: I would maybe remove this dead comment

src/parse/parse_mat.jl
```diff
@@ -0,0 +1,1667 @@
+# CONTRIB file
```
frapac, line 1: I appreciate this `CONTRIB` line!

src/parse/parse_mat.jl
```diff
@@ -0,0 +1,1667 @@
...
+# Originally from PowerModels.jl. Adapted by PowerSystems.jl and by us.
```
frapac, line 2: I would maybe add a link to the original parser in PowerModels.jl

src/powersystem.jl
```diff
@@ -36,42 +35,44 @@ struct PowerNetwork
 
     Sbus::Array{Complex{Float64}}
 
-    function PowerNetwork(datafile::String)
-        data = Parse.parse_raw(datafile)
+    function PowerNetwork(datafile::String, data_format::Int64=0)
```
frapac, line 38: Here, I would be in favor of using multiple dispatch on the `data_format`, and have the following signature there: 
`PowerNetwork(datafile::String, data_format::Format)`

An idea would be to merge the two submodules `ParsePSSE` and `ParseMAT` into a single module `Parser`, and implement in parser:
```julia
abstract type AbstractFormat end

struct MatpowerFormat <: AbstractFormat end
struct PSSEFormat <: AbstractFormat end

```

and then dispatch the parsing on the `AbstractFormat` type. E.g. by replacing
```julia
ParsePSSE.parse_raw(datafile)
ParseMAT.parse_mat(datafile)
```
by 
```julia
data_raw = Parser.parse(MatPowerFormat(), datafile)
data_raw = Parser.parse(PSSEFormat(), datafile)
```

That would allow greater extensibility IMO. But I think this is not blocking for this PR. Just putting this comment here to open the discussion.

src/powersystem.jl
```diff
@@ -240,11 +240,13 @@ Returns vectors indexing buses by type: ref, pv, pq.
...
 function bustypeindex(bus, gen)
     # retrieve indeces
-    BUS_B, BUS_AREA, BUS_VM, BUS_VA, BUS_NVHI, BUS_NVLO, BUS_EVHI,
-    BUS_EVLO, BUS_TYPE = Parse.idx_bus()
+    BUS_I, BUS_TYPE, PD, QD, GS, BS, BUS_AREA, VM, VA, BASE_KV, ZONE, VMAX, VMIN,
+    LAM_P, LAM_Q, MU_VMAX, MU_VMIN = IdxSet.idx_bus()
+    
+    GEN_BUS, PG, QG, QMAX, VG, MBASE, GEN_STATUS, PMAX, PMIN, PC1, PC2, QC1MIN,
+    QC2MIN, QC2MAX, RAMP_AGC, RAMP_10, RAMP_30, RAMP_Q, APF, MU_PMAG, MU_PMIN, MU_QMAX,
+    MU_QMIN = IdxSet.idx_gen()
```
frapac, line 248: I would suggest using a named tuple there, to simplify the management of the elements returned by `idx_gen` :-)

src/powersystem.jl
```diff
@@ -254,8 +256,8 @@ function bustypeindex(bus, gen)
     gencon = zeros(Int8, size(bus, 1))
 
     for i in 1:size(gen, 1)
-        if gen[i, GEN_STAT] == 1
-            gencon[gen[i, GEN_BUS]] += 1
+        if gen[i, GEN_STATUS] == 1
+            gencon[Int(gen[i, GEN_BUS])] += 1
```
frapac, line 260: Why is the array `gen` not of type `AbstracArray{Int}`?

test/perf.jl
```diff
@@ -6,15 +6,22 @@ using LinearAlgebra
 using BenchmarkTools
 using SparseArrays
 using TimerOutputs
-import ExaPF: Parse, PowerSystem
+import ExaPF: ParsePSSE, PowerSystem, IdxSet
```
frapac, line 9: Thanks for updating `test/perf.jl` as well!

test/test_matpower.jl
```diff
@@ -0,0 +1,21 @@
+# Verify solutions against matpower results
```
frapac, line 1: That's nice to have tests implemented for the matpower case!

src/parse/im_common.jl
```diff
@@ -0,0 +1,54 @@
+
+"turns top level arrays into dicts"
+function arrays_to_dicts!(data::Dict{String, <:Any})
+    # update lookup structure
+    for (k, v) in data
+        if isa(v, Array) && length(v) > 0 && isa(v[1], Dict)
+            #println("updating $(k)")
```
frapac, line 7: I would maybe remove this comment

src/parse/parse_mat.jl
```diff
@@ -0,0 +1,1667 @@
...
+        gen_array[i, GEN_BUS] = gen["gen_bus"]
+        gen_array[i, PG] = gen["pg"]
+        gen_array[i, QG] = gen["qg"]
+        gen_array[i, QMAX] = gen["qmax"]
+        gen_array[i, VG] = gen["vg"]
+        gen_array[i, MBASE] = gen["mbase"]
+        gen_array[i, GEN_STATUS] = gen["gen_status"]
+        gen_array[i, PMAX] = gen["pmax"]
+        gen_array[i, PMIN] = gen["pmin"]
+        gen_array[i, PC1] = gen["pc1"]
```
frapac, line 83: Here we could have a small problem. It appears that not all matpower instances have fields `pc1`, `pc2`, `qc1min` and `qc2min`. I do not know if we need this information afterwards. But I think at some point we should add a test to check if the dictionary `gen` has these keys (we could use the function `haskey` for instance)

src/parse/parse_mat.jl
```diff
@@ -0,0 +1,1667 @@
...
+
+    # buses
+    BUS_I, BUS_TYPE, PD, QD, GS, BS, BUS_AREA, VM, VA, BASE_KV, ZONE, VMAX, VMIN,
+    LAM_P, LAM_Q, MU_VMAX, MU_VMIN = IdxSet.idx_bus()
+    
+    nbus = length(data_mat["bus"])
+    bus_array = zeros(nbus, 17)
+
+    for (i, bus) in enumerate(data_mat["bus"])
+        busn = i # we need an external2internal function
```
frapac, line 47: Yes! Today, I learned that almost all matpower instances have their specific IDs for buses. For instance, to import instances from `pglib-opf`, I have to manually build the `external2internal` function when building the adjacency matrix here: https://github.com/exanauts/ExaPF.jl/blob/adrian/matpower/src/powersystem.jl#L378L379
and I added the following lines to import the pglib instances:
```julia
        # build connection matrices
    f = branch[:, F_BUS]                           # list of "from" buses
    t = branch[:, T_BUS]                            # list of "to" buses

    # Convert to valid ordered indexing for matrices
    bus_indexes = convert.(Int, bus[:, BUS_I])
    # in bus_indexes, index shall be unique
    f_i = [findfirst(isequal(el), bus_indexes) for el in f]
    t_i = [findfirst(isequal(el), bus_indexes) for el in t]
```

I do not know if this is the most robust solution, though

src/powersystem.jl
```diff
@@ -331,76 +333,68 @@ end
...
-    st_tr = trans[:, TR_STAT]
-    Ys_tr = st_tr ./ (trans[:, TR_R] .+ im*trans[:, TR_X])
-    tap = (trans[:, TR_WINDV1] ./ trans[:, TR_WINDV2]) .* exp.(im*pi/180 .* trans[:, TR_ANG])
-    Ymag = st_tr .* (trans[:, TR_MAG1] .+ im*trans[:, TR_MAG2])
-    Ytt_tr = Ys_tr
-    Yff_tr = (Ytt_tr ./ (tap .* conj(tap))) .+ Ymag
-    Yft_tr = -Ys_tr ./ conj(tap)
-    Ytf_tr = -Ys_tr ./ tap
-
-    f = [b2i[b] for b in trans[:, TR_FR]]
```
frapac, line 384: why do we remove the `b2i` indexes?

src/ExaPF.jl
dmaldona: Fixed

src/indexes.jl
dmaldona: Yes, the named tuple makes sense. I just followed Matpower's trail and didnt really think this through.

Let me address this in a future PR. I want to focus on get to the RGM by next Tuesday.

src/parse/parse_mat.jl
dmaldona: :)

src/indexes.jl
dmaldona: Let me also say. Using tables for the generator data and such also is not a good idea. I think I might rewrite it and just use a dictionary.

But, again, RGM first.

src/parse/parse_mat.jl
dmaldona: Yes. Done.

src/powersystem.jl
dmaldona: Yes, you're right. This PR is mostly gluing what we had for RAW datafiles and the Matpower format.

While I did not do a clean job I think I've modularized enough so we can refactor this without messing with the rest of the code so we can get back to this after the RGM.

src/powersystem.jl
dmaldona: Yes, 

I think we should use dictionaries instead of what we are doing right now. So:

data["gen"][1]["GEN_BUS"]

which is something similar to what PowerModels.jl has. This, as long as these dicts are not accessed in performance-critical parts of the code.

Again, lets discuss it after we get some sensible results for RGM.

src/powersystem.jl
dmaldona: I am using {Any} array now instead of Floats64. Matpower uses Float64 for the whole array.

I think we should use Dicts instead.

test/perf.jl
dmaldona: :)

test/test_matpower.jl
dmaldona: We now have confirmed power flow has same convergence pattern as matpower. Thats nice.

dmaldona:
@frapac @michel2323 I have updated these PR with your comments and also added the cost function. Once you're OK lets merge.

My plan is to sprint so we can have a RGM prototype by Tuesday. We can note the outstanding issues (e.g. improve the parsed data structures) and write them down as Issues.

src/ExaPF.jl
```diff
@@ -188,6 +192,26 @@ function polar!(Vm, Va, V, ::CUDADevice)
...
+    vmag, vang, pinj, qinj = ExaPF.PowerSystem.retrieve_physics(pf, x, u, p)
+    
+    ref = pf.ref
+    pv = pf.pv
+    pq = pf.pq
+
+    cost = 0.0
+
+    # add pv buses
+    cost += sum(pinj[pv])
```
frapac, line 208: I just see this: here, we have a function named `cost` and a local variable also named `cost`. I would maybe rename one of the two objects to avoid ambiguity.

src/powersystem.jl
```diff
@@ -331,76 +337,67 @@ end
...
+    Cf = sparse(1:nl, f, ones(nl), nl, nb)       # connection matrix for line & from buses
+    Ct = sparse(1:nl, t, ones(nl), nl, nb)       # connection matrix for line & to buses
+
+    # build Yf and Yt such that Yf * V is the vector of complex branch currents injected
+    # at each branch's "from" bus, and Yt is the same for the "to" bus end
+    i = [1:nl; 1:nl]
+    Yf = sparse(i, [f; t], [Yff; Yft], nl, nb)
+    Yt = sparse(i, [f; t], [Ytf; Ytt], nl, nb)
+
+    # Yf = spdiags(Yff, 0, nl, nl) * Cf + spdiags(Yft, 0, nl, nl) * Ct;
```
frapac, line 393: I would maybe remove this dead code

PR #40, opened 2020-09-11
Fix iterative solvers 

### LINKED ISSUE #36: Iterative solvers are broken
It appears that changing the indexing of the Jacobian broke the iterative solvers.

frapac:
This function fixes the iterative solver, **but only on the new API** (I was too lazy to do the job on the old one).  Shall we update the old API as well? 

Fixes #36 

+ this PR also brings minor fixes and cleansing, and add deprecation warning when using functions from the old API.


michel2323:
This is great!

michel2323 (approved):

PR #41, opened 2020-09-11
adjoint of cost function with additions to API

michel2323:
Adjoint cost function that reverses the data flow. Thus the API has to provide adjoint equivalents of the forward computation.


src/PowerSystem/power_network.jl
```diff
@@ -275,6 +275,23 @@ function get_power_injection(fr, v_m, v_a, ybus_re, ybus_im)
     return P
 end
 
+function put_power_injection!(fr, v_m, v_a, adj_v_m, adj_v_a, adj_P, ybus_re, ybus_im)
+    for c in ybus_re.colptr[fr]:ybus_re.colptr[fr+1]-1
+        to = ybus_re.rowval[c]
+        aij = v_a[fr] - v_a[to]
+        # P += v_m[fr]*v_m[to]*(ybus_re.nzval[c]*cos(aij) + ybus_im.nzval[c]*sin(aij))
```
frapac, line 282: could we remove this dead comment?

PR #43, opened 2020-09-16
Remove Jupyter notebooks

frapac:


dmaldona:
@frapac if you're OK with this I can merge.

PR #52, opened 2020-11-06
Benchmarks and Krylov.jl added

michel2323:


benchmark/benchmark.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+# CPU Solvers
+for i in `ls case*` 
+# for i in case300.m
```
frapac, line 5: maybe remove dead comment?

benchmark/benchmark.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+# CPU Solvers
+for i in `ls case*` 
+# for i in case300.m 
+do
+  for j in DirectSolver
+  do
+    julia --project=.. benchmarks.jl $j CPU $i | tail -1 >> cpu.log
```
frapac, line 9: Could we move `julia` executable as a bash variable? That would allow to easily change the Julia version we are using, to benchmark different versions of Julia together :-)

benchmark/benchmarks.jl
```diff
@@ -0,0 +1,76 @@
...
+
+import ExaPF: PowerSystem, LinearSolvers, TimerOutputs
+
+# For debugging in REPL use the following lines
+# empty!(ARGS)
+# push!(ARGS, "BICGSTAB")
+# push!(ARGS, "CUDADevice")
+# push!(ARGS, "caseGO30R-025.raw")
+
+# We do need the time in ms, and not with time units all over the place
```
frapac, line 15: Nice!!

src/LinearSolvers/LinearSolvers.jl
```diff
@@ -164,14 +156,26 @@ function ldiv!(solver::DQGMRES,
...
 )
     P = solver.precond.P
-    @timeit solver.timer "GPU-DQGMRES" begin
-        (y[:], status) = Krylov.dqgmres(J, x, M=P, memory=solver.memory)
-    end
+    (y[:], status) = Krylov.dqgmres(J, x, N=P, memory=solver.memory)
+    return length(status.residuals)
+end
+
+struct KrylovBICGSTAB <: AbstractIterativeLinearSolver
```
frapac, line 163: Nice we now have `KrylovBICGSTAB` in the package

src/LinearSolvers/bicgstab.jl
```diff
@@ -7,7 +7,7 @@ Van der Vorst, Henk A.
 of nonsymmetric linear systems."
 SIAM Journal on scientific and Statistical Computing 13, no. 2 (1992): 631-644.
 """
-function bicgstab(A, b, P, xi, to::TimerOutput;
+function bicgstab(A, b, P, xi;
```
frapac, line 10: indeed, I think it is better to remove `TimerOutput` there. In my observations, it induces a small but non negligible slowdown

frapac (changes requested):
Nice pull request! Overall, I just have a few comments. 
I would suggest moving all instance data in the same folder. Currently, we have some redundancy between:
`test/data`
`benchmark` 

I would move all instances in the same folder. Maybe in `test/data`, or in a new `data` folder. What do you think?

src/LinearSolvers/bicgstab.jl
michel2323: Yes, that was my thought.

benchmark/benchmark.sh
michel2323: I'm not sure since there is no standard environment variable for the Julia executable. This assumes that the environment takes care of that when running the script.

src/LinearSolvers/LinearSolvers.jl
michel2323: The fastest of them all!

benchmark/benchmark.sh
michel2323: > Nice pull request! Overall, I just have a few comments.
> I would suggest moving all instance data in the same folder. Currently, we have some redundancy between:
> `test/data`
> `benchmark`
> 
> I would move all instances in the same folder. Maybe in `test/data`, or in a new `data` folder. What do you think?

Yes. I think a data folder makes sense. Which though raises the question again of making the parser more general as a separate package. But that's for some other time I guess.

benchmark/benchmark.sh
frapac: I was just thinking of setting a local variable at the beginning of the script: 
```bash
JULIA_BIN=julia

for j in DirectSolver
  do
    $JULIA_BIN --project=.. benchmarks.jl $j CPU $i | tail -1 >> cpu.log
```

benchmark/benchmark.sh
michel2323: This fixed by 48bab95. I still kept symbolic links in the benchmarks folder since not every case is benchmarked. Does that make sense @frapac ?

benchmark/benchmark.sh
michel2323: Ok added. Not sure what it changes though.

PR #53, opened 2020-11-06
Evaluators: Implement getter for primal infeasibility 

frapac:


frapac:
Interesting. One test is failing because of a non-deterministic issue. I am unable to reproduce it locally.

michel2323:
I made the initialization of the variables for the sparsity computation of J deterministic. However, it could still fail for some case (though I think very very unlikely). I also added an `@assert` with a comment if such a case arises.

michel2323:
Issue reappeared. I only added more output to gather more data the next time.

PR #55, opened 2020-11-10
Add a wrapper to solve ExaPF problems with MOI compatible solvers

frapac:
This PR brings also some enhancements in the API of `AbstractNLPEvaluators` objects


scripts/moi_solver.jl
```diff
@@ -0,0 +1,43 @@
+
```
michel2323, line 1: Shouldn't this maybe be a test?

scripts/moi_solver.jl
```diff
@@ -0,0 +1,43 @@
...
+    MOI.set(optimizer, MOI.NLPBlock(), block_data)
+    MOI.set(optimizer, MOI.ObjectiveSense(), MOI.MIN_SENSE)
+    MOI.optimize!(optimizer)
+end
+
+datafile = joinpath(dirname(@__FILE__), "..", "data", "case57.m")
+
+nlp = ExaPF.ReducedSpaceEvaluator(datafile)
+aug = ExaPF.AugLagEvaluator(nlp, ExaPF.initial(nlp); c₀=0.1, scale=true)
+# optimizer = KNITRO.Optimizer(hessopt=3)
```
michel2323, line 39: In test you could also add KNITRO.

michel2323 (changes requested):

scripts/moi_solver.jl
frapac: good point!

scripts/moi_solver.jl
frapac: I prefer not to, as it is a commercial solver...

dmaldona (approved):

src/Evaluators/MOI_wrapper.jl
```diff
@@ -0,0 +1,75 @@
+# MOI wrapper
+
+mutable struct ExaEvaluator <: MOI.AbstractNLPEvaluator
```
michel2323, line 3: That sounds now totally picky and maybe unnecessary, but you think we shouldn't call it `ExaPFEvaluator` or maybe even just `Evaluator` like in JuMP?

michel2323 (approved):

src/Evaluators/MOI_wrapper.jl
frapac: No, this is particularly relevant. Indeed, I should be more careful with naming. I have renamed it `MOIEvaluator`, to emphasis the fact that we rely on MOI there.

PR #62, opened 2020-11-18
Pre-release cleansings 

frapac:
- clean-up `Project.toml` and adds compatible versions for all deps packages
- add default tolerance for `Krylov.bicgstab` 
- change default behavior for Newton-Krylov: when algorithm failed to converge, issues a warning instead of an error
- minor cleanings in the documentation & add documentation for `LinearSolvers.ldiv!` 
- remove `TimerOutputs` in submodule `LinearSolvers` (and only in this submodule)

### DIFF
--- a/src/models/polar/kernels.jl
+++ b/src/models/polar/kernels.jl
@@ -1,11 +1,11 @@
 # Implement kernels for polar formulation
 
 """
-residualFunction
+power_balance
 
 Assembly residual function for N-R power flow
 """
-function residualFunction(V, Ybus, Sbus, pv, pq)
+function power_balance(V, Ybus, Sbus, pv, pq)
     # form mismatch vector
     mis = V .* conj(Ybus * V) - Sbus
     # form residual vector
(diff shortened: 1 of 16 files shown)


src/models/polar/kernels.jl
```diff
@@ -1,11 +1,11 @@
 # Implement kernels for polar formulation
 
 """
-residualFunction
+power_balance
 
 Assembly residual function for N-R power flow
 """
-function residualFunction(V, Ybus, Sbus, pv, pq)
+function power_balance(V, Ybus, Sbus, pv, pq)
```
michel2323, line 8: Why did you change this?

frapac: Good question. Even if this is an external function, this function is being used only in the tests (as it is not compatible with GPU). But, as we renamed `residualFunction_polar` to `residual_polar!`, I thought that it would be more consistent to rename `residualFunction` as well.

PR #63, opened 2020-11-23
increased tolerance

michel2323:
This test fails on Summit so I increased the tolerance. I am not sure though whether that is correct or maybe we want to take a more general look at the tolerances.


michel2323:
More context:
```
Power flow 300 bus case: Test Failed at /autofs/nccs-svm1_proj/csc359/michel/ExaPF.jl/test/matpower.jl:108
Expression: isapprox(norm(cache.balance, Inf), 1.3783e-12, rtol = 0.01)
Evaluated: isapprox(1.4479528687161292e-12, 1.3783e-12; rtol = 0.01)
```

test/matpower.jl
```diff
@@ -105,7 +105,7 @@ end
     ExaPF.get!(polar, State(), x, cache)
 
     @test conv.n_iterations == 5
-    @test isapprox(norm(cache.balance, Inf), 1.3783e-12, rtol=1e-2)
+    @test isapprox(norm(cache.balance, Inf), 1.3783e-12, rtol=1e-1)
```
frapac, line 108: Mmmmh. thinking more about it, the residual should be close to 0 at the end. At the moment we are comparing with the final value we get with matpower, but I think it would make more sense to compare it with 0: 
`@test isapprox(norm(cache.balance, Inf), 0, rtol=1e-1)`
Not sure that `rtol` makes any sense for numerical values this small :-)

dmaldona: @michel2323 @frapac this makes sense (comparing with 0).

Its however intriguing, that the final residual value seemed to be quite stable/robust along different platforms (my laptop, desktop, moonshot...).

michel2323: Fixed according to @frapac 's suggestions. @dmaldona Summit uses PPC CPUs. So this is the first system with a different architecture and instruction set.

PR #65, opened 2020-12-01
Remove timer reset

michel2323:
This removes the timer reset in the NR. The reset destroys all the timers and makes them inaccessible once `ExaPF.update!()` returns. I needed that changed in the benchmarks and I think it also makes sense. The user has to manually reset the timers via `ExaPF.reset_timer!(ExaPF.TIMER)`.


frapac:
My bad, I added the reset  in a previous PR... 
I thought it would make more sense if the timing was attached to one Newton-Raphson solve (especially if we are running the code on different instances, with different sizes, and on various architectures). But what you said make also sense, in a benchmark perspective.

frapac:
Also, instead of merging, could we rebase this branch w.r.t. `release` to avoid an additional merge commit?

PR #74, opened 2020-12-11
Optionally run only powerflow on device

michel2323:
We want to be able to run only the powerflow (Newton-Raphson) on the GPU with the entire optimization stack on the CPU.


src/Evaluators/reduced_evaluator.jl
```diff
@@ -86,6 +86,43 @@ function ReducedSpaceEvaluator(
     )
 end
 
+function ReducedSpaceEvaluator(nlp::ReducedSpaceEvaluator, device = nothing) 
+    if device == CUDADevice()
+        VT = CuVector 
+        MT = CuMatrix
+    elseif device == CPU()
+        buffer = PolarNetworkState(nlp.buffer, device)
```
frapac, line 94: Do we really need to instantiate the `buffer` here? I am wondering whether this line is redundant with line 101.

src/Evaluators/reduced_evaluator.jl
```diff
@@ -86,6 +86,43 @@ function ReducedSpaceEvaluator(
     )
 end
 
+function ReducedSpaceEvaluator(nlp::ReducedSpaceEvaluator, device = nothing) 
+    if device == CUDADevice()
```
frapac, line 90: On a more generic standpoint (i.e. not necessarily related to this PR) I am wondering whether we should use our own types to handle the different devices, instead of KA's ones. Indeed, I am not sure how we could specify that we want to use both CPU and GPU in KA. An idea of implementation could be: 
```julia
abstract type AbstractExaDevice end

struct CPU <: AbstractExaDevice end

abstract type AbstractGPUDevice <: AbstractExaDevice end
struct ExaCUDA <: AbstractGPUDevice end
struct ExaROCm <: AbstractGPUDevice end
struct ExaOneAPI <: AbstractGPUDevice end

abstract type ExaMixedDevice <: AbstractExaDevice end
struct MixedGPUDevice <: ExaMixedDevice end

```
(just thinking out loud, I think we could improve this abstraction)

src/Evaluators/reduced_evaluator.jl
```diff
@@ -86,6 +86,43 @@ function ReducedSpaceEvaluator(
     )
 end
 
+function ReducedSpaceEvaluator(nlp::ReducedSpaceEvaluator, device = nothing)
```
frapac, line 89: do we want to use `device=nothing` as default?

src/Evaluators/reduced_evaluator.jl
```diff
@@ -119,14 +156,21 @@ initial(nlp::ReducedSpaceEvaluator) = initial(nlp.model, Control())
...
 bounds(nlp::ReducedSpaceEvaluator, ::Constraints) = nlp.g_min, nlp.g_max
 
-function update!(nlp::ReducedSpaceEvaluator, u; verbose_level=0)
+function update!(nlp::ReducedSpaceEvaluator, u; verbose_level=0, device = nothing)
     x₀ = nlp.x
     jac_x = nlp.autodiff.Jgₓ
     # Transfer x, u, p into the network cache
     transfer!(nlp.model, nlp.buffer, nlp.x, u, nlp.p)
     # Get corresponding point on the manifold
+    if !(device === nothing)
```
frapac, line 165: I would rather use multiple dispatch there, to avoid this `if` statement

src/Evaluators/reduced_evaluator.jl
```diff
@@ -119,14 +156,21 @@ initial(nlp::ReducedSpaceEvaluator) = initial(nlp.model, Control())
...
+        jac_x = AutoDiff.StateJacobian(jac_x, device)
+        nlp  = ReducedSpaceEvaluator(nlp, device)
```
frapac, line 167: Do we know how long does it takes to instantiate on the GPU a new `ReducedSpaceEvaluator` from `nlp`, on the largest instances?

src/autodiff.jl
```diff
@@ -66,86 +66,127 @@ struct StateJacobian{VI, VT, MT, SMT, VP, VD, SubT, SubD} <: AbstractJacobian
...
+    VD = typeof(t1sx)
+    return StateJacobian{VI, VT, MT, SMT, VP, VD, typeof(varx), typeof(t1svarx)}(
+        J, compressedJ, coloring, t1sseeds, t1sF, x, t1sx, map, varx, t1svarx
+    )
+end
+
+function StateJacobian(jac::StateJacobian, device)
+    ncolor = size(unique(jac.coloring),1)
+    t1s{N} = ForwardDiff.Dual{Nothing,Float64, N} where N
+    if device == CUDADevice()
```
frapac, line 155: Again, not related to this PR, but I think in the long term we should find a way to automatically dispatch the proper Julia types depending on the device where we are deporting the computation upon.

frapac (commented):
Looks good overall! I just have a few comments, mostly about the design of the code. 
They are mostly on how to handle properly the different devices in ExaPF, and are not related to this PR in particular. But I think it's time to foster this discussion.

src/Evaluators/reduced_evaluator.jl
michel2323: That's a mistake :-)

src/Evaluators/reduced_evaluator.jl
michel2323: I don't know. It's in a way to force the device in the powerflow. IMO neither `CPU()` nor `CUDADevice()` is right.

src/Evaluators/reduced_evaluator.jl
michel2323: For sure. The initialization has to be moved. I'm wondering how to best do this such that one can essentially instantiate the evaluator for a device only if it has not been instantiated before. One evaluator can be on the CPU, CUDA, AMDGPU, or oneAPI.

michel2323: What we could do is to instintiate it for every device type (aka SHADOW COPY), but that may be overkill.

src/autodiff.jl
michel2323: Yes, it's the CSR CSC issue. we'd have to introduce an `AbstractSparseMatrix` type. We could also here handle it such that we instantiate it for every device type available.

src/Evaluators/reduced_evaluator.jl
michel2323: Yes. Though we should maybe wait for KA to do the first step. But not sure...

src/Evaluators/reduced_evaluator.jl
michel2323: This is resolved now.

src/Evaluators/reduced_evaluator.jl
michel2323: Done.

src/Evaluators/common.jl
```diff
@@ -8,6 +8,12 @@ struct AutoDiffFactory <: AbstractAutoDiffFactory
     ∇f::AdjointStackObjective
 end
 
+function transfer!(target::AutoDiffFactory, origin::AutoDiffFactory)
```
frapac, line 11: How about overloading `Base.copy!` there?

src/Evaluators/reduced_evaluator.jl
frapac: I will open a separate issue to discuss this

src/autodiff.jl
```diff
@@ -38,6 +38,17 @@ function _init_seed!(t1sseeds, coloring, ncolor, nmap)
     end
 end
 
+function transfer!(target::AbstractJacobian, origin::AbstractJacobian)
```
frapac, line 41: Idem: how about renaming this function `Base.copy!`?

src/autodiff.jl
```diff
@@ -38,6 +38,17 @@ function _init_seed!(t1sseeds, coloring, ncolor, nmap)
...
+    if target.x isa Array
```
frapac, line 42: I think we could use multiple dispatch there. But IMO, it would make more sense to update the code once the PR about AD refactoring is merged

src/models/polar/adjoints.jl
```diff
@@ -17,6 +17,16 @@ struct AdjointStackObjective{VT}
     jvᵤ::VT
 end
 
+function transfer!(target::AdjointStackObjective, origin::AdjointStackObjective)
```
frapac, line 20: `Base.copy!` ?

src/models/caches.jl
```diff
@@ -22,3 +22,23 @@ struct PolarNetworkState{VT} <: AbstractNetworkBuffer
     dx::VT
 end
 
+function PolarNetworkState(state::PolarNetworkState, device = nothing)
+    if device == CUDADevice()
+        VT = CuVector{Float64}
+    elseif device == CPU()
+        VT = Vector{Float64}
+    else
+        VT = typeof(state.vmag)
```
frapac, line 31: Here, I think we should find a better way to infer the type rather than using this hardcoded line (but I confess I am quite guilty of this in other parts of the code...)

frapac: I think it would make sense to gather all comments porting on the type inference in a separate issue, that we could discuss together

test/powerflow.jl
```diff
@@ -37,4 +37,25 @@ import ExaPF: PowerSystem
             @test convergence.n_iterations == 2
         end
     end
+    @testset "Deport only powerflow on device $device" for device in DEVICES
```
frapac, line 40: Not sure we want to test all devices there, but I have to think more about that.

test/powerflow.jl
```diff
@@ -37,4 +37,25 @@ import ExaPF: PowerSystem
...
+        uk = ExaPF.initial(polar, Control())
+        p = ExaPF.initial(polar, Parameters())
+
+        @testset "Powerflow solver $(LinSolver)" for LinSolver in [ExaPF.LinearSolvers.DirectSolver]
+            algo = LinSolver(precond)
+            xk = copy(x0)
+            nlp = ExaPF.ReducedSpaceEvaluator(polar, xk, uk, p;
+                                              ε_tol=tolerance, linear_solver=algo)
+            nlp_pf = ExaPF.ReducedSpaceEvaluator(nlp, device)
+            convergence = ExaPF.update!(nlp, uk; verbose_level=ExaPF.VERBOSE_LEVEL_NONE, nlp_pf = nlp_pf)
```
frapac, line 55: I am not 100% sure with this API, but I have no better idea to propose at the moment. I think we could discuss on how to improve the API after merging this PR.

frapac:
Before merging this PR, I would suggest doing a rebase to get rid of the `fix` and `Merge` commits. What do you think ?

test/powerflow.jl
michel2323: > Before merging this PR, I would suggest doing a rebase to get rid of the `fix` and `Merge` commits. What do you think ?
Of course we do rebase 😊

michel2323:
> Before merging this PR, I would suggest doing a rebase to get rid of the `fix` and `Merge` commits. What do you think ?

Of course we do rebase 😊

src/autodiff.jl
michel2323: It's not really a copy though. Only the relevant parts for the computation are transferred. Not everything is copied like the mapping.

test/powerflow.jl
michel2323: Until it breaks... why not? In the end we want to transfer from any device to any device if needed.

src/Evaluators/common.jl
michel2323: Same as below... it's not really a copy. it assumes both were initialized with the same case.

frapac:
I took some time to think more about this PR. I think we could do better than the current solution, for the `update!` function. Let me explain further my thoughts. 

- With #82 , we have a single buffer to consider in the `ReducedSpaceEvaluator`. This buffer is used both for the powerflow resolution, and for the evaluations of the objective and the different constraints. 
- If we want to deport on the GPU only the computation of the powerflow, we would need two buffers: one defined in the host memory, the other one defined on the device (to run the powerflow equation).
- each time we have a new control `u` to consider, we should transfer its values to both buffers. 

By doing that, we would need to duplicate only the buffer, instead of duplicating the evaluator (current solution, with `nlp` and `nlp_pf`). 

We should just think carefully about the implementation. Ideally, we should have only one buffer if we are using only the CPU or the GPU.

PR #75, opened 2020-12-14
Refactoring AutoDiff

michel2323:
* Same constructor and evaluation functions for state Jacobian and control Jacobian.
* Move indexing and sparsity to `model.jl`. This needs follow-up work.

### DIFF
--- a/src/autodiff.jl
+++ b/src/autodiff.jl
@@ -54,7 +54,8 @@ Creates an object for the Jacobian
 * `varx::SubT`: View of `map` on `x`
 * `t1svarx::SubD`: Active (AD) view of `map` on `x`
 """
-struct Jacobian{VI, VT, MT, SMT, VP, VD, SubT, SubD} <: AbstractJacobian
+# struct Jacobian{VI, VT, MT, SMT, VP, VD, SubT, SubD} <: AbstractJacobian
+struct Jacobian{VI, VT, MT, SMT, VP, VD, SubT, SubD} 
     J::SMT
     compressedJ::MT
     coloring::VI
@@ -66,9 +67,11 @@ struct Jacobian{VI, VT, MT, SMT, VP, VD, SubT, SubD} <: AbstractJacobian
     # Cache views on x and its dual vector to avoid reallocating on the GPU
     varx::SubT
     t1svarx::SubD
-    function Jacobian(structure, F, v_m, v_a, ybus_re, ybus_im, pinj, qinj, pv, pq, ref, nbus)
-        nv_m = size(v_m, 1)
-        nv_a = size(v_a, 1)
+    function Jacobian(structure, F, v_m, v_a, ybus_re, ybus_im, pinj, qinj, pv, pq, ref, nbus, type)
+        nv_m = length(v_m)
+        nv_a = length(v_a)
+        npbus = length(pinj)
+        nref = length(ref)
         if F isa Array
             VI = Vector{Int}
             VT = Vector{Float64}
@@ -85,14 +88,8 @@ struct Jacobian{VI, VT, MT, SMT, VP, VD, SubT, SubD} <: AbstractJacobian
             error("Wrong array type ", typeof(F))
         end
 
-        mappv = [i + nv_m for i in pv]
-        mappq = [i + nv_m for i in pq]
-        # Ordering for x is (θ_pv, θ_pq, v_pq)
         map = VI(structure.map)
-        nmap = size(map,1)
-
-        # Used for sparsity detection with randomized inputs
-
+        nmap = length(structure.map)
         # Need a host arrays for the sparsity detection below
         spmap = Vector(map)
         hybus_re = Spmat{Vector{Int}, Vector{Float64}}(ybus_re)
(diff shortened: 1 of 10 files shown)


test/powersystem.jl
```diff
@@ -101,16 +101,22 @@ const PS = PowerSystem
...
         F = zeros(Float64, npv + 2*npq)
         # Compute Jacobian at point V manually and use it as reference
-        J = ExaPF.residual_jacobian(V, Ybus, pv, pq)
+        J = residual_jacobian(V, Ybus, ref, pv, pq)
         J♯ = copy(J)
+        # Build the Jacobian structure
+        mappv = [i + npv for i in pv]
+        mappq = [i + npv for i in pq]
+        # Ordering for x is (θ_pv, θ_pq, v_pq)
+        statemap = vcat(mappv, mappq, pq)
```
michel2323, line 110: @frapac This is the test that is failing. It seems `pv` `pq` have a different indexing than we do internally in ExaPF, at least at the stage when AutoDiff is called.  Since that map has to be created outside of AutoDiff, I tried to use the same logic than internally in ExaPF but this fails. The resulting maps look different. Are the indices reshuffled somewhere in `PolarForm` when they come from PS?

src/autodiff.jl
```diff
@@ -54,7 +54,8 @@ Creates an object for the state Jacobian
 * `varx::SubT`: View of `map` on `x`
 * `t1svarx::SubD`: Active (AD) view of `map` on `x`
 """
-struct StateJacobian{VI, VT, MT, SMT, VP, VD, SubT, SubD} <: AbstractJacobian
+# struct Jacobian{VI, VT, MT, SMT, VP, VD, SubT, SubD} <: AbstractJacobian
```
frapac, line 57: maybe remove dead comment?

src/autodiff.jl
```diff
@@ -122,130 +102,34 @@ struct StateJacobian{VI, VT, MT, SMT, VP, VD, SubT, SubD} <: AbstractJacobian
         Vre = Float64.([i for i in 1:n])
         Vim = Float64.([i for i in n+1:2*n])
         V = Vre .+ 1im .* Vim
-        J = residual_jacobian(V, Y, pv, pq)
+        J = structure.sparsity(V, Y, ref, pv, pq)
```
frapac, line 105: Good point. 
I am wondering if we should not pass directly the matrix `J` as input for the sparsity detection phase. I am afraid that at the moment we are not completely generic, as we could not infer a priori the signature of the function `structure.sparsity`

src/autodiff.jl
```diff
@@ -122,130 +102,34 @@ struct StateJacobian{VI, VT, MT, SMT, VP, VD, SubT, SubD} <: AbstractJacobian
...
-            diagIbus    = sparse(1:n, 1:n, Ibus, n, n)
-            diagVnorm   = sparse(1:n, 1:n, V./abs.(V), n, n)
-
-            dSbus_dVm = diagV * conj(Ybus * diagVnorm) + conj(diagIbus) * diagVnorm
-            dSbus_dpbus = diagV * conj(Ybus * diagVnorm) + conj(diagIbus) * diagVnorm
-
-            j11 = real(dSbus_dVm[[pv; pq], [ref; pv; pv]])
-            j21 = imag(dSbus_dVm[pq, [ref; pv; pv]])
-            J = [j11; j21]
+        if isa(type, ControlJacobian)
```
frapac, line 122: `else if`?

src/autodiff.jl
```diff
@@ -254,6 +138,9 @@ struct ControlJacobian{VI, VT, MT, SMT, VP, VD, SubT, SubD} <: AbstractJacobian
     end
 end
 
+struct StateJacobian <: AbstractJacobian end
```
frapac, line 141: Could we move the definition of these two structures right after line 25 and the definition of `AbstractJacobian`? To ensure that we define first the types, then we proceed to the implementation.

frapac: I would maybe go further, and use instead directly the `State` and `Control` type introduced in `src/model/model.jl`: 
https://github.com/exanauts/ExaPF.jl/blob/master/src/models/models.jl#L34-L50

src/models/polar/kernels.jl
```diff
@@ -32,7 +32,7 @@ balance function [`power_balance`](@ref).
 # Note
 Code adapted from MATPOWER.
 """
-function residual_jacobian(V, Ybus, pv, pq)
+function residual_jacobian(V, Ybus, ref, pv, pq)
```
frapac, line 35: True. We should add `ref` to the signature of this function, indeed

src/models/polar/polar.jl
```diff
@@ -300,8 +360,11 @@ function powerflow(
         iter += 1
 
         @timeit TIMER "Jacobian" begin
-            AutoDiff.residual_jacobian!(jacobian, residual_polar!, Vm, Va,
-                                   polar.ybus_re, polar.ybus_im, pbus, qbus, pv, pq, ref, nbus, TIMER)
+            AutoDiff.residual_jacobian!(jacobian, residual_polar!, 
+                                   Vm, Va,
+                                   polar.ybus_re, polar.ybus_im, pbus, qbus, pv, pq, ref, nbus, AutoDiff.StateJacobian())
+            # AutoDiff.residual_jacobian!(jacobian, residual_polar!, Vm, Va,
```
frapac, line 366: I would maybe remove dead comment, there?

frapac (commented):
That's a really nice work, the code looks a lot cleaner than before! I think this PR is ready to merge once we figure out how to fix the test breaking currently.

src/autodiff.jl
michel2323: I agree. I should look into that. I put it on my list. That's nearly there.

src/autodiff.jl
michel2323: 😬postponed.

PR #77, opened 2020-12-16
Add TagBot Github Actions

frapac:


michel2323:
Into master?

frapac:
Indeed, I have hesitated about that... But as we are doing the release in master directly, I thought that it would make more sense to merge this PR in master.

Issue #57, opened 2020-11-12
Implement a new structure NewtonRaphson to store options of the algorithm


frapac:
As we did in the `LinearSolvers` submodule, it may be interesting to implement a `NewtonRaphson` structure to store in one place all the options of the algorithm, and ease the implementation of other non-linear algorithms (as the decoupled formulation introduced in the article [Fast decoupled flows](https://ieeexplore.ieee.org/abstract/document/4075431/)).

A prototype could be: 
```julia
abstract type AbstractNonLinearSolver end 

struct NewtonRaphson <: AbstractNonLinearSolver 
    linear_solver::AbstractLinearSolver # direct or indirect 
    tolerance::Float64
    max_iter::Int
end
```
and we could dispatch the resolution of the powerflow equations with:
```julia
powerflow(
    polar::PolarForm, 
    buffer::PolarNetworkState,
    algo::NewtonRaphson,
)
```

frapac:
Closed by #88

PR #88, opened 2021-01-06
Add a NewtonRaphson object storing all arguments for the powerflow solver

frapac:
I think that would allow greater modularity. It would also help implementing other powerflow algorithm (e.g. `DecoupledLoadFlow` :smile: ), as we could now dispatch on the algorithm we are using.


frapac:
Solve https://github.com/exanauts/ExaPF.jl/issues/57

michel2323 (approved):

PR #90, opened 2021-01-13
Sparse Hessian for Powerflow

michel2323:


michel2323:
Was solved with #99  avoiding forward over forward. We ran into performance issues with KA and second-order ForwardDiff.

Issue #92, opened 2021-01-20
Change indexing of control variable


frapac:
Currently, the control `u` is indexed as: 
`u = [vmag[ref]; pg[pv]; vmag[pv]]`
I believe it would be more consistent to order the voltage together, such as
`u = [vmag[ref]; vmag[pv]; pg[pv]]`. 
That would avoid ugly expressions like 
![image](https://user-images.githubusercontent.com/15050705/105214784-bceb9880-5b50-11eb-8f9e-dfd9ca368733.png)
and simplify some part of the code.

frapac:
Closed with #93

PR #93, opened 2021-01-20
Implement reduced Hessian on CPU

frapac:
* compute reduced Hessian using the adjoint-adjoint method, using MATPOWER expressions
* correctness of reduced Hessian expressions is checked with finite differences, in `test/hessians.jl` 
* AutoDiff routines remain to be implemented. Once done, we will be able to evaluate the Hessian on the GPU 

*Warning* 
* side effect: change slightly the structure of ExaPF, by introducing a new template file `src/models.jl`


frapac:
I have changed the ordering of `u` in this PR, in 961e712.
New ordering is: 
`u = [vmag[ref]; vmag[pv]; pg[pv]]` 
see #92

michel2323 (approved):

PR #99, opened 2021-02-12
Hessian adjoint tangent projection using ForwardDiff over adjoint residual function

michel2323:


src/autodiff.jl
```diff
@@ -382,6 +382,136 @@ function residual_jacobian!(arrays::Jacobian,
...
+                             v_m, v_a, ybus_re, ybus_im, pinj, qinj, pv, pq, ref, nbus,
+                             type::AbstractJacobian)
+    nvbus = length(v_m)
+    ninj = length(pinj)
+    F = zeros(Float64, length(arrays.t1sF))
+    adj_F = zeros(Float64, length(arrays.t1sF))
+    if isa(type, StateJacobian)
+        arrays.x[1:nvbus] .= v_m
+        arrays.x[nvbus+1:2*nvbus] .= v_a
+        global dJ = zeros(Float64, length(F), length(arrays.map))
```
frapac, line 396: Why use `global` here?

frapac: I assume that's because of the scope of the variables defined in the `if` statement. I think it would be more appropriate to define `dJ` before line 393, with
```
local dJ
```

src/autodiff.jl
```diff
@@ -382,6 +382,136 @@ function residual_jacobian!(arrays::Jacobian,
     return nothing
 end
 
+function residual_jacobian_adj!(arrays::Jacobian,
```
frapac, line 385: I would maybe add a comment explaining that we are computing the full Jacobian using `m` calls to the adjoint code `residual_adj_polar!`

frapac: And also that `dJ` will be evaluated as a dense matrix first.

src/autodiff.jl
```diff
@@ -382,6 +382,136 @@ function residual_jacobian!(arrays::Jacobian,
...
+        x[nvbus+1:nvbus+ninj] .= pinj
+        t1sx .= arrays.x
+        adj_t1sx .= 0.0
+        t1sF .= 0.0
+        adj_t1sF .= 0.0
+    else
+        error("Unsupported Jacobian structure")
+    end
+    t1sF .= 0.0
+    adj_t1sF .= lambda
```
frapac, line 474: Nice!!

src/polar/kernels.jl
```diff
@@ -113,6 +113,95 @@ function residual_polar!(F, v_m, v_a,
...
+        coef_cos = v_m[fr]*v_m[to]*ybus_re_nzval[c]
+        coef_sin = v_m[fr]*v_m[to]*ybus_im_nzval[c]
+
+        cos_val = cos(aij)
+        sin_val = sin(aij)
+        F[i] += coef_cos * cos_val + coef_sin * sin_val
+        if i > npv
+            F[npq + i] += coef_cos * sin_val - coef_sin * cos_val
+        end
+        if i > npv
```
frapac, line 148: These three lines are redundant with the three lines before?

src/polar/kernels.jl
```diff
@@ -113,6 +113,95 @@ function residual_polar!(F, v_m, v_a,
...
+    # adj_pinj[fr] -= adj_F[i]
+end
+
+function residual_adj_polar!(F, adj_F, v_m, adj_v_m, v_a, adj_v_a,
+                         ybus_re, ybus_im,
+                         pinj, adj_pinj, qinj, adj_qinj, pv, pq, nbus)
+    npv = length(pv)
+    npq = length(pq)
+    if isa(F, Array)
+        kernel! = residual_adj_kernel!(KA.CPU(), 4)
```
frapac, line 193: Why using 4 CPU threads here?

test/powersystem.jl
```diff
@@ -120,6 +120,10 @@ const PS = PowerSystem
             jacobianAD, ExaPF.residual_polar!, Vm, Va,
             ybus_re, ybus_im, pbus, qbus, pv, pq, ref, nbus, ExaPF.AutoDiff.StateJacobian())
         @test jacobianAD.J ≈ J♯
+        J = ExaPF.AutoDiff.residual_jacobian_adj!(
```
frapac, line 123: I would maybe move this code in `test/Polar/gradient.jl`

frapac (commented):
I like it!

src/polar/kernels.jl
```diff
@@ -113,6 +113,95 @@ function residual_polar!(F, v_m, v_a,
...
+    else
+        kernel! = residual_adj_kernel!(KA.CUDADevice(), 256)
+    end
+    ev = kernel!(F, adj_F, v_m, adj_v_m, v_a, adj_v_a,
+                 ybus_re.nzval, ybus_re.colptr, ybus_re.rowval,
+                 ybus_im.nzval, ybus_im.colptr, ybus_im.rowval,
+                 pinj, adj_pinj, qinj, adj_qinj, pv, pq, nbus,
+                 ndrange=npv+npq)
+    wait(ev)
```
frapac, line 202: I would add a new line right-after: 
```
return nothing
```
To ensure that the compiler could be inferred properly the return type of this function

src/autodiff.jl
```diff
@@ -382,6 +382,136 @@ function residual_jacobian!(arrays::Jacobian,
...
+            pv, pq, nbus
+        )
+        ps = ForwardDiff.partials.(adj_t1sx[arrays.map])
+        for i in 1:size(arrays.J,1)
+            arrays.compressedJ[:,i] .= ps[i].values
+        end
+    else
+        error("Unsupported Jacobian structure")
+    end
+    uncompress_kernel!(arrays.J, arrays.compressedJ, arrays.coloring)
```
frapac, line 512: We have an issue here: the type of `arrays.J` is inferred as `SparseMatrixCSC`, instead of `SparseMatrixCSC{Float64, Int}`. That leads to some inefficiencies in `uncompress_kernel!`

frapac: In fact, on line 80 of `src/autodiff.jl`, we should replace: 
```
 SMT = SparseMatrixCSC
```julia
by
```julia
 SMT = SparseMatrixCSC{Float64, Int}
```
(yeah, type inference is really my new dada)

frapac (commented):
For some reason `AutoDiff.residual_hessian!` is allocating a lot on large cases. This should not be the case. I am currently investigating that.

src/autodiff.jl
```diff
@@ -382,6 +382,136 @@ function residual_jacobian!(arrays::Jacobian,
...
+    seed_kernel!(arrays.t1sseeds, arrays.varx, arrays.t1svarx, nbus)
+    if isa(type, StateJacobian)
+        residual_adj_polar!(
+            t1sF, adj_t1sF,
+            view(t1sx, 1:nvbus),
+            view(adj_t1sx, 1:nvbus),
+            view(t1sx, nvbus+1:2*nvbus),
+            view(adj_t1sx, nvbus+1:2*nvbus),
+            ybus_re, ybus_im,
+            pinj, zeros(eltype(pinj), length(pinj)),
```
frapac, line 485: Why instantiating a new array there?

michel2323: Those are the adjoints. They have to be zero.

michel2323:
Sorry @frapac, a lot rewiring. In the end it's much less code. I removed everything we don't need. That's also why I might want to keep the history this time. In case I want to do that seeding again.

I did a merge back from develop and resolved a few minor conflicts.

src/polar/kernels.jl
michel2323: No idea. That 4 must have emerged at some point.

src/autodiff.jl
```diff
@@ -116,7 +116,7 @@ struct Jacobian{VI, VT, MT, SMT, VP, VD, SubT, SubD}
         end
         t1s{N} = ForwardDiff.Dual{Nothing,Float64, N} where N
         if isa(type, StateJacobian)
-            x = VT(zeros(Float64, nv_m + nv_a))
+            x = VT(zeros(Float64, nvbus + nvbus))
```
frapac, line 119: Maybe replace with `2 * nvbus`?

src/autodiff.jl
```diff
@@ -145,6 +145,69 @@ struct Jacobian{VI, VT, MT, SMT, VP, VD, SubT, SubD}
...
+    t1svarx::SubD
+    function Hessian(structure, F, vm, va, ybus_re, ybus_im, pinj, qinj, pv, pq, ref)
+        nvbus = length(vm)
+        npbus = length(pinj)
+        nref = length(ref)
+        if F isa Array
+            VI = Vector{Int}
+            VT = Vector{Float64}
+            MT = Matrix{Float64}
+            SMT = SparseMatrixCSC
```
frapac, line 178: I would specify explicitly 
`SMT = SparseMatrixCSC{Float64, Int}`
for the sake of type inference

src/autodiff.jl
```diff
@@ -145,6 +145,69 @@ struct Jacobian{VI, VT, MT, SMT, VP, VD, SubT, SubD}
...
+            A = Vector
+        elseif F isa CUDA.CuArray
+            VI = CUDA.CuVector{Int}
+            VT = CUDA.CuVector{Float64}
+            MT = CUDA.CuMatrix{Float64}
+            SMT = CUSPARSE.CuSparseMatrixCSR
```
frapac, line 184: Idem

src/autodiff.jl
```diff
@@ -145,6 +145,69 @@ struct Jacobian{VI, VT, MT, SMT, VP, VD, SubT, SubD}
...
+            A = CUDA.CuVector
+        else
+            error("Wrong array type ", typeof(F))
+        end
+
+        map = VI(structure.map)
+        nmap = length(structure.map)
+        # Need a host arrays for the sparsity detection below
```
frapac, line 192: I would maybe remove this comment, as we are not computing any sparsity detection below (as far as I understand)?

src/autodiff.jl
```diff
@@ -319,68 +382,135 @@ function uncompress_kernel!(J::CUSPARSE.CuSparseMatrixCSR, compressedJ, coloring
...
+* `J::Jacobian`: Factory created Jacobian object to update
 * `residual_polar`: Primal function
-* `v_m, v_a, ybus_re, ybus_im, pinj, qinj, pv, pq, ref, nbus`: Inputs both
+* `vm, va, ybus_re, ybus_im, pinj, qinj, pv, pq, ref, nbus`: Inputs both
   active and passive parameters. Active inputs are mapped to `x` via the preallocated views.
-
+* `type::AbstractJacobian`: Either `StateJacobian` or `ControlJacobian`
 """
-function residual_jacobian!(arrays::Jacobian,
+function residual_jacobian!(J::Jacobian,
```
frapac, line 398: This renaming is nice!

src/autodiff.jl
```diff
@@ -319,68 +382,135 @@ function uncompress_kernel!(J::CUSPARSE.CuSparseMatrixCSR, compressedJ, coloring
...
 end
 
+"""
+    residual_hessian_adj_tgt!(H::Hessian,
+                        residual_adj_polar!,
+                        lambda, tgt,
+                        vm, va, ybus_re, ybus_im, pinj, qinj, pv, pq, ref,
+                        nbus)
+
+Update the sparse Jacobian entries using AutoDiff. No allocations are taking place in this function.
```
frapac, line 455: is it true? cf line 485: 
```julia
t1sseedvec = zeros(Float64, length(x))
```

src/autodiff.jl
```diff
@@ -319,68 +382,135 @@ function uncompress_kernel!(J::CUSPARSE.CuSparseMatrixCSR, compressedJ, coloring
...
+    x[nvbus+1:2*nvbus] .= v_a
+    x[2*nvbus+1:2*nvbus+ninj] .= pinj
+    t1sx .= H.x
+    adj_t1sx .= 0.0
+    t1sF .= 0.0
+    adj_t1sF .= lambda
+    # Seeding
+    nmap = length(H.map)
+    t1sseedvec = zeros(Float64, length(x))
+    for i in 1:nmap
```
frapac, line 486: I would maybe add an `@inbounds` macro there

src/autodiff.jl
```diff
@@ -319,68 +382,135 @@ function uncompress_kernel!(J::CUSPARSE.CuSparseMatrixCSR, compressedJ, coloring
...
+        t1sseedvec .= 0.0
+    end
+    seed_kernel!(H.t1sseeds, H.varx, H.t1svarx, nbus)
+    residual_adj_polar!(
+        t1sF, adj_t1sF,
+        view(t1sx, 1:nvbus), view(adj_t1sx, 1:nvbus),
+        view(t1sx, nvbus+1:2*nvbus), view(adj_t1sx, nvbus+1:2*nvbus),
+        ybus_re, ybus_im,
+        view(t1sx, 2*nvbus+1:2*nvbus+ninj), view(adj_t1sx, 2*nvbus+1:2*nvbus+ninj),
+        qinj, zeros(eltype(qinj), length(qinj)),
```
frapac, line 498: here, we are allocating an array with size `length(qinj)`. Is this expected?

src/polar/kernels.jl
```diff
@@ -113,6 +113,95 @@ function residual_polar!(F, v_m, v_a,
...
+        coef_sin = v_m[fr]*v_m[to]*ybus_im_nzval[c]
+
+        cos_val = cos(aij)
+        sin_val = sin(aij)
+        F[i] += coef_cos * cos_val + coef_sin * sin_val
+        if i > npv
+            F[npq + i] += coef_cos * sin_val - coef_sin * cos_val
+        end
+        if i > npv
+            F[npq + i] += coef_cos * sin_val - coef_sin * cos_val
```
frapac, line 149: I don't understand the following code. It looks like to me that this line is redundant with the line 146 above?

src/polar/kernels.jl
```diff
@@ -113,6 +113,95 @@ function residual_polar!(F, v_m, v_a,
...
+            adj_sin_val  +=  coef_cos * adj_F[npq + i]
+        end
+
+        adj_coef_cos +=  cos_val  * adj_F[i]
+        adj_coef_sin +=  sin_val  * adj_F[i]
+        adj_cos_val  +=  coef_cos * adj_F[i]
+        adj_sin_val  +=  coef_sin * adj_F[i]
+
+        adj_aij =   cos(aij)*adj_sin_val
+        adj_aij += -sin(aij)*adj_cos_val
```
frapac, line 170: we have computed `{cos,sin}(aij)` before, on line 142: 
```julia
        cos_val = cos(aij)
        sin_val = sin(aij)
```
I would suggest using these values here, as `cos` and `sin` are expensive operations.

src/polar/kernels.jl
```diff
@@ -113,6 +113,95 @@ function residual_polar!(F, v_m, v_a,
...
+        adj_v_m[fr] += v_m[to]*ybus_im_nzval[c]*adj_coef_sin
+        adj_v_m[to] += v_m[fr]*ybus_im_nzval[c]*adj_coef_sin
+        adj_v_m[fr] += v_m[to]*ybus_re_nzval[c]*adj_coef_cos
+        adj_v_m[to] += v_m[fr]*ybus_re_nzval[c]*adj_coef_cos
+
+        adj_v_a[fr] += adj_aij
+        adj_v_a[to] -= adj_aij
+    end
+    # qinj is not active
+    # if i > npv
```
frapac, line 181: Maybe remove dead comment? Or explain why we are keeping it from now?

frapac (commented):
Nice work! I am quite convinced by this adjoint-tangent approach. I made a few comments, porting mostly on minor details.

src/autodiff.jl
```diff
@@ -145,6 +145,69 @@ struct Jacobian{VI, VT, MT, SMT, VP, VD, SubT, SubD}
...
+
+        map = VI(structure.map)
+        nmap = length(structure.map)
+        # Need a host arrays for the sparsity detection below
+        if F isa CUDA.CuArray
+            J = CUSPARSE.CuSparseMatrixCSR(J)
+        end
+        t1s{N} = ForwardDiff.Dual{Nothing,Float64, N} where N
+        x = VT(zeros(Float64, 2*nvbus + npbus))
+        t1sx = A{t1s{1}}(x)
```
frapac, line 198: I just did a pass on type inference. Here, `t1sx` because `A` is a generic type `Array` (on the CPU), whereas it should be inferred as a `Array{Dual{Nothing, Float64, 1}}`

test/powersystem.jl
michel2323: This is gone now.

src/autodiff.jl
michel2323: This is gone.

src/autodiff.jl
michel2323: No `m` calls anymore. This code is gone too.

src/autodiff.jl
michel2323: Yes, this is sweet!

src/polar/kernels.jl
michel2323: I am not sure. I think with KA that is fine. We do the same with `residual_polar!`.

src/autodiff.jl
michel2323: Done.

src/autodiff.jl
michel2323: Yea yea done :-).

src/autodiff.jl
michel2323: Idem

src/autodiff.jl
michel2323: Thanks. I think this was one of the oldest untouched code sections. It still needs more work.

src/autodiff.jl
michel2323: Yes. And the creation of `J` was also dead code.

src/autodiff.jl
michel2323: fixed

src/autodiff.jl
michel2323: Ok. The bounds checking actually creates a real overhead. So I'm not sure whether to use it everywhere or not.

src/autodiff.jl
michel2323: I think so. These are the adjoint of `qinj` which there are also `length(qinj)`.

michel2323: I revise this. You are right, we never compute the adjoint of `qinj`, so removed it.

src/polar/kernels.jl
michel2323: Oops. Fixed.

src/polar/kernels.jl
michel2323: Fixed.

src/autodiff.jl
michel2323: We should look into all of this. Also the seeding. I agree, let's to that in a performance specific PR.

src/autodiff.jl
frapac: In fact, even if bounds checking creates an overhead, it is actually negligible on modern architecture, as far as I observed

src/polar/kernels.jl
frapac: Ok then!

frapac (approved):

PR #100, opened 2021-02-16
Implement Hessian-vector product in MOI wrapper

frapac:
- allows to use Hv product in ExaTron and Knitro
- Hessians in full-space are cached to avoid recomputing them each time we compute a `Hv` prod
- The structure `HessianFactory` should be adapted once #99 is merged 
- implement a new evaluator to compute a feasible point (`FeasibleEvaluator`) 
- also, fix type inference in `MOIEvaluator`

### DIFF
--- a/src/Evaluators/common.jl
+++ b/src/Evaluators/common.jl
@@ -1,3 +1,18 @@
+# Common interface for AbstractNLPEvaluator
+#
+function Base.show(io::IO, nlp::AbstractNLPEvaluator)
+    n = n_variables(nlp)
+    m = n_constraints(nlp)
+    println(io, "A Evaluator object")
+    println(io, "    * #vars: ", n)
+    println(io, "    * #cons: ", m)
+end
+
+function constraint(nlp::AbstractNLPEvaluator, x)
+    cons = similar(x, n_constraints(nlp)) ; fill!(cons, 0)
+    constraint!(nlp, cons, x)
+    return cons
+end
 
 function gradient(nlp::AbstractNLPEvaluator, x)
     ∇f = similar(x) ; fill!(∇f, 0)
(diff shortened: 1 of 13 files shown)


src/Evaluators/common.jl
```diff
@@ -50,6 +85,17 @@ struct AutoDiffFactory{VT} <: AbstractAutoDiffFactory
     ∇f::AdjointStackObjective{VT}
 end
 
+struct HessianFactory{SpMT}
+    hashu::UInt64
```
michel2323, line 89: What is this?

src/Evaluators/reduced_evaluator.jl
```diff
@@ -278,17 +285,36 @@ function jacobian!(nlp::ReducedSpaceEvaluator, jac, u)
...
+function jprod!(nlp::ReducedSpaceEvaluator, jv, u, v)
+    nᵤ = length(u)
+    m  = n_constraints(nlp)
+    @assert nᵤ == length(v)
+
+    # jprod! is an expensive operation in the reduced space,
+    # as we need to evaluate the full reduced Jacobian. In case we will
+    # call jprod! several times, we cache the reduced Jacobian
+    # for future use.
+    if isnothing(nlp.jacobian_stack) || nlp.jacobian_stack.hashu != hash(u)
```
michel2323, line 309: I see :-) . What `u` is on the GPU? Is this okay?

src/Evaluators/common.jl
frapac: It just stores the hash for the last variable `u` evaluated. By doing that, we avoid computing several times the Hessian for a same variable `u` (when evaluating hessian vector product). That should be removed once the new AD backend for Hessian is implemented.

src/Evaluators/reduced_evaluator.jl
frapac: At the moment, we could not evaluate Hessian and Jacobian on the GPU.

michel2323: Given the size of `u` it might be better to just store a copy and check for equivalence?

michel2323 (approved):

src/Evaluators/common.jl
michel2323: Got it.

PR #101, opened 2021-02-17
Update documentation

frapac:
Improve consistency of the API
* PS.Generator -> PS.Generators
* state_constraint -> state_constraints

### DIFF
--- a/docs/src/man/evaluators.md
+++ b/docs/src/man/evaluators.md
@@ -4,78 +4,95 @@ In `ExaPF.jl`, the evaluators are the final layer of the structure.
 They take as input a given `AbstractFormulation` and implement the
 callbacks for the optimization solvers.
 
-## Overview
+## Overview of the AbstractNLPEvaluator
 
-An `AbstractEvaluator` is tailored to the optimization problem
+An `AbstractNLPEvaluator` implements an optimization problem
 associated with an underlying `AbstractFormulation`:
 ```math
 \begin{aligned}
-\min_{x, u} \;          & f(x, u)     \\
-\text{subject to} \quad & g(x, u) = 0 \\
-                        & h(x, u) \leq 0.
+\min_{u \in \mathbb{R}^n} \;             & f(u)     \\
+\text{subject to} \quad & g(u) = 0 \\
+                        & h(u) \leq 0.
 \end{aligned}
 ```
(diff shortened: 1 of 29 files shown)


docs/make.jl
```diff
@@ -1,8 +1,8 @@
 using Pkg
 
-Pkg.develop(PackageSpec(path=joinpath(dirname(@__FILE__), "..")))
+# Pkg.develop(PackageSpec(path=joinpath(dirname(@__FILE__), "..")))
 # when first running instantiate
-Pkg.instantiate()
+# Pkg.instantiate()
```
michel2323, line 5: These can be removed for good as `build_docs.jl` does that indeed.

docs/src/man/evaluators.md
```diff
@@ -124,23 +141,27 @@ To juggle between the mathematical description (characterized
 by a state $x$ and a control $u$) and the physical description (characterized
 by the voltage and power injection at each bus), the evaluator `nlp`
 stores internally a cache `nlp.buffer`, with type `AbstractNetworkBuffer`.
+```julia-repl
+julia> buffer = get(nlp, ExaPF.PhysicalState())
+```
 
 #### Evaluation of the callbacks
 
 Now that we have a `nlp` evaluator available, we could embed it in any
```
michel2323, line 150: Be aware that "we could" is like french "on pourrait". I think you want to say "we can". Also it may be better to use passive tense. "Now that the `nlp` evaluator is available, it can be embedded in any optimization routine".  This theme is recurring a few times.

docs/src/man/evaluators.md
```diff
@@ -4,78 +4,95 @@ In `ExaPF.jl`, the evaluators are the final layer of the structure.
...
+julia> ExaPF.constraint!(nlp, cons, u)
 
-## A journey in the reduced space
+```
 
-When we aim at optimizing the problem directly in the reduced
-space manifold, the `ReducedSpaceNLPEvaluator` is our workhorse.
-We recall that the reduced space is defined implicitly by the
+
+## A journey in the reduced space with the ReducedSpaceEvaluator
```
michel2323, line 64: That was a nice trip!

michel2323 (approved):

docs/make.jl
frapac: Sure! Sorry, I should have removed that before doing the PR

PR #102, opened 2021-02-17
Fix the performance of the seeding by moving it onto the host

michel2323:


src/autodiff.jl
```diff
@@ -7,10 +7,12 @@ using CUDA
 import CUDA.CUSPARSE
 import ForwardDiff
 import SparseDiffTools
+using KernelAbstractions
 
 using ..ExaPF: Spmat, xzeros, State, Control
 
 import Base: show
+KA = KernelAbstractions
```
frapac, line 15: I would change this line as:
`const KA = KernelAbstractions`

src/autodiff.jl
```diff
@@ -28,17 +30,15 @@ struct StateJacobian <: AbstractJacobian end
...
-function _init_seed!(t1sseeds, coloring, ncolor, nmap)
-    t1sseedvec = zeros(Float64, ncolor)
-    @inbounds for i in 1:nmap
-        for j in 1:ncolor
-            if coloring[i] == j
-                t1sseedvec[j] = 1.0
-            end
+KA.@kernel function _init_seed!(t1sseeds, t1sseedvecs, coloring, ncolor)
+    i = @index(Global, Linear)
+    t1sseedvecs[:,i] .= 0
```
frapac, line 35: oh, I see! `t1seedvecs` is now a 2d array. What size will it take on memory for bigger case? Also, I think we could initiate `t1sseedvecs` to 0.0 before launching the kernel, if each index `i` is affected to a particular row

src/autodiff.jl
```diff
@@ -115,32 +119,21 @@ struct Jacobian{VI, VT, MT, SMT, VP, VD, SubT, SubD}
...
-            t1svarx = view(t1sx, map)
-        else
-            error("Unsupported Jacobian type. Must be either ControlJacobian or StateJacobian.")
-        end
-
-        VP = typeof(t1sseeds)
-        VD = typeof(t1sx)
-        return new{VI, VT, MT, SMT, VP, VD, typeof(varx), typeof(t1svarx)}(
-            J, compressedJ, coloring, t1sseeds, t1sF, x, t1sx, map, varx, t1svarx
+        init_seed_kernel! = _init_seed!(KA.CPU(), 4)
```
frapac, line 122: I would maybe add a comment that we always do the seeding on the CPU

frapac (approved):
Nice that it was so quick to fix! I just have minor comments on this PR.

src/autodiff.jl
michel2323: The same than the compressed Jacobian. So ncolors x output dimension.

src/autodiff.jl
michel2323: Done.

src/autodiff.jl
michel2323: Done.

src/autodiff.jl
michel2323: It shouldn't be a problem. Plus after this call it should be gc'ed.

PR #103, opened 2021-02-18
Parallel CPU and improved GPU preconditioner

michel2323:
* Set default number of threads for CPU KA to `Threads.nthreads()`
* GPU preconditioner is now one kernel per step instead of `nblock` kernels
* CPU preconditioner now also uses threads

This still requires some regression testing for performance. I hope it's an improvement.

### DIFF
--- a/src/LinearSolvers/preconditioners.jl
+++ b/src/LinearSolvers/preconditioners.jl
@@ -1,3 +1,4 @@
+using KernelAbstractions
 """
     AbstractPreconditioner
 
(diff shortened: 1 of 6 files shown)


michel2323:
This is also supposed to solve the issue with non-deterministic results on the GPU.

src/LinearSolvers/preconditioners.jl
```diff
@@ -23,22 +24,45 @@ Creates an object for the block-Jacobi preconditioner
...
-    map
-    cumap
-    part
-    cupart
-    P
-    id
-    function BlockJacobiPreconditioner(J, npart, device=KA.CPU())
-        if isa(J, CUSPARSE.CuSparseMatrixCSR)
+struct BlockJacobiPreconditioner{AT,GAT,VI,GVI,MT,GMT,MI,GMI,SMT} <: AbstractPreconditioner
+    nblocks::Int64
```
frapac, line 28: The best practice is to use `Int` instead of `Int64` (for 32 bits architecture)

src/LinearSolvers/preconditioners.jl
```diff
@@ -55,18 +79,28 @@ mutable struct BlockJacobiPreconditioner <: AbstractPreconditioner
         for (i,v) in enumerate(part)
             push!(partitions[v], i)
         end
-        Js = Vector{Matrix{Float64}}(undef, npart)
-        nJs = maximum(length.(partitions))
-        id = Matrix{Float64}(I, nJs, nJs)
+        lpartitions = VI(undef, npart)
```
frapac, line 82: I would replace this line with broadcasting:
`lpartitions = length.(partitions)`

src/LinearSolvers/preconditioners.jl
```diff
@@ -55,18 +79,28 @@ mutable struct BlockJacobiPreconditioner <: AbstractPreconditioner
...
+        for i in 1:npart
+            lpartitions[i] = length(partitions[i])
+        end
+        blocksize = maximum(length.(partitions))
```
frapac, line 86: In fact, you did the broadcasting here :-) I would rephrase this line as 
`blocksize = maximum(lpartitions)`
to avoid doing twice the same operation

src/LinearSolvers/preconditioners.jl
```diff
@@ -55,18 +79,28 @@ mutable struct BlockJacobiPreconditioner <: AbstractPreconditioner
...
+        blocks = AT(undef, blocksize, blocksize, npart)
```
frapac, line 87: I would be careful there. E.g. 
```julia
a = rand(100)
VT = CuVector{Float64}
@time VT(a);
  0.000097 seconds (10 allocations: 240 bytes)

@time convert(VT, a);
  0.000084 seconds (7 allocations: 192 bytes)
```
It's usually better to use `convert` rather than `VT(...)`.

frapac: Also, if the type is already set to the destination:
```julia
VT = Vector{Float64}
@time convert(VT, a);
  0.000005 seconds
@time VT(a);
  0.000006 seconds (1 allocation: 896 bytes)
```
One version is allocating, the other not

frapac (approved):
Nice that now `BlockJacobiPreconditioner` is properly typed!

src/LinearSolvers/preconditioners.jl
michel2323: lol psst.

src/LinearSolvers/preconditioners.jl
michel2323: I tried to dig into the source code, because I don't understand the difference. First, if I rerun the code I get this:
```julia

julia> @time VT(a)
  0.000088 seconds (9 allocations: 224 bytes)

julia> @time convert(VT, a)
  0.000082 seconds (9 allocations: 224 bytes)
```
But that's Julia 1.5.3.

src/LinearSolvers/preconditioners.jl
michel2323: I spent too much time in code converting `int` to `long long int` due to overflows that I tend to be on the cautious side. In fact, I don't see a reason for not being explicit. The code would still work on a 32bit architecture. And if it's indeed a performance issue we should explicitly have an Integer type.

michel2323: Which we could actually also do for the `Float64`, especially on the GPU.

PR #106, opened 2021-02-24
Generic AutoDiff for constraints

frapac:
Allow to apply AutoDiff to compute the Jacobian of any constraints. 
The AutoDiff for Jacobian is moved inside `src/Polar`

In addition: 
- `power_balance` (previously `residual_polar`) is now defined as a generic (equality) constraint. 
- we have renamed the constraints, for clarity: 
```
state_constraints -> voltage_magnitude_constraints
power_constraints -> {active_power_constraints,reactive_power_constraints}
```
- we have created a new directory `src/Polar/Constraints` to define all the constraints


michel2323:
I think `src/Polar/autodiff.jl` may just be called `derivatives.jl`. Eventually, ExaPF should be unaware whether AD is used or not. And the more we move forward, the more I think that AutoDiff might become just a package for AD on GPUs of unstructured problems in optimization independent of ExaPF.

In fact anything in the folder Polar would in fact be provided by a modeling framework that @youngdae  develops.

michel2323 (approved):

PR #107, opened 2021-02-25
Handwritten adjoints and removal of Zygote

michel2323:
This PR introduces handwritten adjoints for all the constraints. The line flow constraint adjoints exhibit the same issues than the Zygote ones: They disagree with ForwardDiff and FD about `1e-4`. This issue has to be kept open. It is either a bug or some unknown numerical issue.

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -20,6 +20,8 @@ jobs:
       - uses: julia-actions/julia-runtest@latest
 
   test-moonshot:
+    env:
+      CUDA_VISIBLE_DEVICES: 1
     runs-on: self-hosted
     strategy:
       matrix:
--- a/src/Polar/Constraints/line_flow.jl
+++ b/src/Polar/Constraints/line_flow.jl
@@ -24,46 +24,19 @@ function flow_constraints(polar::PolarForm, cons, vm, va, pbus, qbus)
 end
 
 function flow_constraints_grad!(polar::PolarForm, cons_grad, buffer, weights)
-    T = typeof(buffer.vmag)
-    if isa(buffer.vmag, Array)
-        kernel! = accumulate_view!(KA.CPU())
-    else
-        kernel! = accumulate_view!(KA.CUDADevice())
-    end
-
     nlines = PS.get(polar.network, PS.NumberOfLines())
     nbus = PS.get(polar.network, PS.NumberOfBuses())
-    f = polar.topology.f_buses
-    t = polar.topology.t_buses
-    # Statements references below (1)
-    fr_vmag = buffer.vmag[f]
-    to_vmag = buffer.vmag[t]
-    fr_vang = buffer.vang[f]
-    to_vang = buffer.vang[t]
-    function lumping(fr_vmag, to_vmag, fr_vang, to_vang)
-        Δθ = fr_vang .- to_vang
-        cosθ = cos.(Δθ)
-        sinθ = sin.(Δθ)
-        cons = branch_flow_kernel_zygote(
-            polar.topology.yff_re, polar.topology.yft_re, polar.topology.ytf_re, polar.topology.ytt_re,
-            polar.topology.yff_im, polar.topology.yft_im, polar.topology.ytf_im, polar.topology.ytt_im,
-            fr_vmag, to_vmag,
-            cosθ, sinθ
-        )
-        return dot(weights, cons)
-    end
-    grad = Zygote.gradient(lumping, fr_vmag, to_vmag, fr_vang, to_vang)
-    gvmag = @view cons_grad[1:nbus]
-    gvang = @view cons_grad[nbus+1:2*nbus]
-    # This is basically the adjoint of the statements above (1). = becomes +=.
-    ev_vmag = kernel!(gvmag, grad[1], f, ndrange = nbus)
-    ev_vang = kernel!(gvang, grad[3], f, ndrange = nbus)
-    wait(ev_vmag)
-    wait(ev_vang)
-    ev_vmag = kernel!(gvmag, grad[2], t, ndrange = nbus)
-    ev_vang = kernel!(gvang, grad[4], t, ndrange = nbus)
-    wait(ev_vmag)
-    wait(ev_vang)
+    PT = polar.topology
+    fill!(cons_grad, 0)
+    adj_vmag = @view cons_grad[1:nbus]
+    adj_vang = @view cons_grad[nbus+1:2*nbus]
+    ev = adj_branch_flow_kernel!(polar.device)(weights, buffer.vmag, adj_vmag, 
+            buffer.vang, adj_vang,
+            PT.yff_re, PT.yft_re, PT.ytf_re, PT.ytt_re,
+            PT.yff_im, PT.yft_im, PT.ytf_im, PT.ytt_im,
+            PT.f_buses, PT.t_buses, nlines, ndrange = nlines
+    )
+    wait(ev)
     return cons_grad
 end
 
--- a/src/Polar/kernels.jl
+++ b/src/Polar/kernels.jl
@@ -588,7 +588,6 @@ KA.@kernel function adj_reactive_power_edge_kernel!(
 
     # Reverse run
     adj_inj = adj_qg[i_gen]
-    adj_qload[bus] = adj_qg[i_gen]
     @inbounds for c in ybus_re_colptr[bus]:ybus_re_colptr[bus+1]-1
         to = ybus_re_rowval[c]
         aij = vang[bus] - vang[to]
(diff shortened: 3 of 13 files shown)


src/polar/kernels.jl
```diff
@@ -244,6 +307,52 @@ function transfer!(polar::PolarForm, buffer::PolarNetworkState, u)
     wait(ev)
 end
 
+KA.@kernel function adj_transfer_kernel!(
+    adj_vmag, adj_vang, adj_pinj, adj_qinj, adj_u, pv, pq, ref, adj_pload, adj_qload
```
frapac, line 311: I guess here `adj_pload` is the adjoint of the power generation `pg` (`adj_pg`), as the load is constant?

src/polar/kernels.jl
```diff
@@ -130,14 +133,14 @@ KA.@kernel function residual_adj_kernel!(F, adj_F, v_m, adj_v_m, v_a, adj_v_a,
     if i > npv
         F[i + npq] -= qinj[fr]
     end
-    @inbounds for c in ybus_re_colptr[fr]:ybus_re_colptr[fr+1]-1
+    @inbounds for c in colptr[fr]:colptr[fr+1]-1
```
frapac, line 136: Nice cleansing!

src/polar/kernels.jl
```diff
@@ -181,22 +192,74 @@ KA.@kernel function residual_adj_kernel!(F, adj_F, v_m, adj_v_m, v_a, adj_v_a,
...
+
+function adj_residual_polar!(F, adj_F, vm, adj_vm, va, adj_va,
                          ybus_re, ybus_im,
                          pinj, adj_pinj, qinj, pv, pq, nbus)
     npv = length(pv)
     npq = length(pq)
+    nvbus = length(vm)
+    nnz = length(ybus_re.nzval)
+    T = eltype(vm)
+    edge_vm_from = zeros(T, nnz)
```
frapac, line 226: Do you know if we could reuse these variables to compute the adjoint of the line flow constraints?

frapac (commented):
Just some preliminary comments on this PR. I need to write down the math to better understand what we are doing :grimacing:

src/polar/kernels.jl
michel2323: Yes we could. I guess this belongs into the cache together with the adjoints.

michel2323:
> Just some preliminary comments on this PR. I need to write down the math to better understand what we are doing 😬

I take shortcuts in the code; like for example when to use `+=` etc. This is not the best example to learn manual differentiation.

src/polar/kernels.jl
michel2323: I am not sure. In that case we could remove that argument. I am not yet sure how to handle the `adjoint!` signatures in general. What we also have to clarify is weither the adjoint routines return a primal value. I would not recommend that.

.github/workflows/action.yml
```diff
@@ -20,6 +20,8 @@ jobs:
       - uses: julia-actions/julia-runtest@latest
 
   test-moonshot:
+    env:
+      CUDA_VISIBLE_DEVICES: 1
```
frapac, line 24: Why setting this value? (I am curious :smiley_cat: )

src/Polar/Constraints/line_flow.jl
```diff
@@ -24,46 +24,25 @@ function flow_constraints(polar::PolarForm, cons, vm, va, pbus, qbus)
 end
 
 function flow_constraints_grad!(polar::PolarForm, cons_grad, buffer, weights)
-    T = typeof(buffer.vmag)
     if isa(buffer.vmag, Array)
-        kernel! = accumulate_view!(KA.CPU())
+        kernel! = adj_branch_flow_kernel!(KA.CPU())
```
frapac, line 28: When `polar` is available in the scope, I would suggest defining the kernel directly as
```julia
kernel! = adj_branch_flow_kernel!(polar.device)
```

src/Polar/kernels.jl
```diff
@@ -1,4 +1,5 @@
 import KernelAbstractions: @index
+using ForwardDiff
```
frapac, line 2: I don't think we need to import `ForwardDiff` here, as it is already loaded in `src/ExaPF.jl`

src/Polar/kernels.jl
```diff
@@ -404,51 +747,70 @@ KA.@kernel function branch_flow_kernel!(
...
-end
+    # branch apparent power limits - from bus
+    yff_abs = yff_re[ℓ]^2 + yff_im[ℓ]^2
+    yft_abs = yft_re[ℓ]^2 + yft_im[ℓ]^2
+    yre_fr =   yff_re[ℓ] * yft_re[ℓ] + yff_im[ℓ] * yft_im[ℓ]
+    yim_fr = - yff_re[ℓ] * yft_im[ℓ] + yff_im[ℓ] * yft_re[ℓ]
 
-"""
-    accumulate_view(x, vx, indices)
+    # not needed in the reverse run
```
frapac, line 770: Do we need this dead comment?

src/Polar/kernels.jl
```diff
@@ -404,51 +747,70 @@ KA.@kernel function branch_flow_kernel!(
...
-KA.@kernel function accumulate_view!(x, vx, indices)
-    # This is parallelizable
-    id = @index(Global, Linear)
-    for (j, idx) in enumerate(indices)
-        if id == idx
-            x[id] += vx[j]
-        end
-    end
+    # not needed in the reverse run
+    # to_flow = vmag[to_bus]^2 * (
```
frapac, line 784: and this one?

src/Polar/kernels.jl
```diff
@@ -404,51 +747,70 @@ KA.@kernel function branch_flow_kernel!(
...
+    #     ytf_abs * vmag[fr_bus]^2 + ytt_abs * vmag[to_bus]^2 +
+    #     2 * vmag[fr_bus] * vmag[to_bus] * (yre_to * cosθ - yim_to * sinθ)
+    # )
+    # slines[ℓ + nlines] = to_flow
+
+    adj_to_flow = adj_slines[ℓ + nlines]
+    adj_vmag[to_bus] += (2 * vmag[to_bus] * ytf_abs * vmag[fr_bus]^2
```
frapac, line 791: I would replace `2` by `2.0` (same `4` and `6`)

src/Polar/kernels.jl
```diff
@@ -404,51 +747,70 @@ KA.@kernel function branch_flow_kernel!(
...
+    adj_vmag[to_bus] += (2 * yft_abs * vmag[fr_bus]^2 * vmag[to_bus]
+                       + 2 * vmag[fr_bus]^3 * (yre_fr * cosθ - yim_fr * sinθ)
+                        ) * adj_from_flow
+    adj_cosθ += 2.0 * vmag[to_bus] * vmag[fr_bus]^3 *   yre_fr  * adj_from_flow
+    adj_sinθ += 2.0 * vmag[to_bus] * vmag[fr_bus]^3 * (-yim_fr) * adj_from_flow
+
+    adj_Δθ =   cosθ * adj_sinθ
+    adj_Δθ -=  sinθ * adj_cosθ
+    adj_vang[fr_bus] += adj_Δθ
+    adj_vang[to_bus] -= adj_Δθ
```
frapac, line 815: Nice job! This adjoint does not look trivial

frapac (commented):
Ok to merge once minor comments have been addressed

src/Polar/kernels.jl
michel2323: That's why it's off!

src/Polar/kernels.jl
michel2323: You are right. I tried that actually yesterday to make sure that's not the cause of the numerical issues.

src/Polar/kernels.jl
michel2323: I prefer to have the forward code as a comment in the adjoint function. That's just a preference, we can remove it if you don't like it. It's easier for readability of the adjoint code if someone looks at it.

src/Polar/kernels.jl
michel2323: ok!

src/Polar/Constraints/line_flow.jl
michel2323: oh nice! indeed!

src/Polar/kernels.jl
frapac: Ok to keep the comment, then!

.github/workflows/action.yml
michel2323: This makes the tests on moonshot run on `device=1` to avoid a clash with most users and #110

PR #109, opened 2021-02-26
Implement adjoint! function 

frapac:
- add a new function `adjoint!` to compute the adjoint of the constraints
- refactor `AutoDiff.hessian!` to use `adjoint!`
- fix evaluation of adjoint for `reactive_power_constraints`

### DIFF
--- a/src/Polar/kernels.jl
+++ b/src/Polar/kernels.jl
@@ -588,7 +588,6 @@ KA.@kernel function adj_reactive_power_edge_kernel!(
 
     # Reverse run
     adj_inj = adj_qg[i_gen]
-    adj_qload[bus] = adj_qg[i_gen]
     @inbounds for c in ybus_re_colptr[bus]:ybus_re_colptr[bus+1]-1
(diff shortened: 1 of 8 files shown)


src/Polar/kernels.jl
```diff
@@ -604,8 +603,8 @@ KA.@kernel function adj_reactive_power_edge_kernel!(
         adj_coef_sin = -cos_val  * adj_inj
         adj_cos_val  = -coef_sin * adj_inj
 
-        adj_aij =   cos_val*adj_cos_val
-        adj_aij += -sin_val*adj_sin_val
+        adj_aij =   coef_cos * cos_val * adj_inj
+        adj_aij +=  coef_sin * sin_val * adj_inj
```
michel2323, line 607: I think you could change this to
```
        adj_aij =   cos_val*adj_sin_val
        adj_aij += -sin_val*adj_cos_va
```
That would be like an AD machine would write it.

michel2323 (approved):

PR #112, opened 2021-03-01
polar: add generic jtprod evaluation

frapac:
- fix bug in computation of the adjoint
- refactor the computation of the objective
- allow to compute the Hessian of the objective using the adjoint of active_power_constraints


src/Polar/kernels.jl
```diff
@@ -299,47 +299,48 @@ function transfer!(polar::PolarForm, buffer::PolarNetworkState, u)
...
 
 # Transfer values in (x, u) to buffer
-function adj_transfer!(polar::PolarForm, buffer::PolarNetworkState, adj_u)
-    if isa(u, CUDA.CuArray)
-        kernel! = adj_transfer_kernel!(KA.CUDADevice())
-    else
-        kernel! = adj_transfer_kernel!(KA.CPU())
-    end
-    nbus = length(buffer.vmag)
+function adjoint_transfer!(
```
frapac, line 330: `adjoint_transfer!` implements the adjoint of the `transfer!` function once and for all. We are computing both the adjoints w.r.t. `x` and w.r.t. `u` here.

michel2323 (approved):

PR #114, opened 2021-03-02
Adjoint of residual, reactive power, and line flows on the GPU using KA

michel2323:


src/Polar/kernels.jl
```diff
@@ -1,6 +1,7 @@
 import KernelAbstractions: @index
+using CUDA
+using ForwardDiff
```
frapac, line 3: Do we need to import `ForwardDiff` explicitly here? I think `ForwardDiff` is already present in this scope

src/Polar/kernels.jl
```diff
@@ -204,31 +205,92 @@ KA.@kernel function adj_residual_node_kernel!(F, adj_F, vm, adj_vm, va, adj_va,
...
+                                  colptr, rowval,
+                                  edge_vm_from, edge_va_from) 
+    i = @index(Global, Linear)
+    @inbounds for c in colptr[i]:colptr[i+1]-1
+        adj_vm[i] += edge_vm_from[c]
+        adj_va[i] += edge_va_from[c]
+    end
+end
+
+KA.@kernel function adj_residual_node_kernel2!(adj_vm, adj_va,
```
frapac, line 218: Interesting. Is `adj_residual_node_kernel2!` faster than `adj_residual_node_kernel1!`?

frapac: In fact, I am not sure to understand what exactly the purpose of `adj_residual_node_kernel2` is

src/Polar/kernels.jl
```diff
@@ -204,31 +205,92 @@ KA.@kernel function adj_residual_node_kernel!(F, adj_F, vm, adj_vm, va, adj_va,
...
+    ev = kernel_node1!(adj_vm, adj_va,
+            ybus_re.colptr, ybus_re.rowval,
+            edge_vm_from, edge_va_from, 
+            ndrange=nvbus)
+    wait(ev)
+    ev = kernel_node2!(adj_vm, adj_va,
+            tspedge_vm_to.nzVal, tspedge_vm_to.rowInd, tspedge_vm_to.colInd,
+            tspedge_va_to.nzVal, tspedge_va_to.rowInd, tspedge_va_to.colInd,
+            ndrange=nnz)
+    wait(ev)
```
frapac, line 272: Do you know whether we could merge `kernel_node1!` and `kernel_node2!` ?

src/Polar/kernels.jl
```diff
@@ -204,31 +205,92 @@ KA.@kernel function adj_residual_node_kernel!(F, adj_F, vm, adj_vm, va, adj_va,
...
+    edge_vm_from = CUDA.zeros(T, nnz)
+    edge_vm_to   = CUDA.zeros(T, nnz)
+    edge_va_from = CUDA.zeros(T, nnz)
+    edge_va_to   = CUDA.zeros(T, nnz)
+    kernel_edge! = adj_residual_edge_kernel!(KA.CUDADevice())
+    kernel_node1! = adj_residual_node_kernel1!(KA.CUDADevice())
+    kernel_node2! = adj_residual_node_kernel2!(KA.CUDADevice())
+    spedge_vm_to = CUSPARSE.CuSparseMatrixCSR{T}(colptr, rowval, edge_vm_to,(nvbus,nvbus))
+    spedge_va_to = CUSPARSE.CuSparseMatrixCSR{T}(colptr, rowval, edge_va_to,(nvbus,nvbus))
+    tspedge_vm_to = CUSPARSE.CuSparseMatrixCOO(spedge_vm_to)
```
frapac, line 251: I see. Do we use `COO` there to get the `(I, J, V)` representation?

test/Polar/hessian.jl
```diff
@@ -70,8 +73,8 @@ const PS = PowerSystem
         ExaPF.∂cost(polar, ∂obj, cache)
         ∇fₓ = ∂obj.∇fₓ
         ∇fᵤ = ∂obj.∇fᵤ
-        λ  = -(∇gₓ') \ ∇fₓ
-        grad_adjoint = ∇fᵤ + ∇gᵤ' * λ
+        λ  = -(Array(∇gₓ')) \ Array(∇fₓ)
+        grad_adjoint = Array(∇fᵤ) + Array(∇gᵤ)' * λ
```
frapac, line 77: My bad ... I should improve the tests, so it would be easier to port them on the GPU...

src/Polar/kernels.jl
michel2323: kernel1 adds the from-edge to the nodes and goes over all buses. kernel2 goes over all the to-edges and is a bit trickier, but now I realize it may still be wrong. It should also go over all the buses. Not sure why each thread gets a line.

src/Polar/kernels.jl
michel2323: Could be. Let's look at that now.

src/Polar/kernels.jl
michel2323: Correct. And that conversion only touches the array I. J and V stay in place. I should use `convert` here.

test/Polar/hessian.jl
michel2323: I think I solved it better than the above now :-) .

src/Polar/kernels.jl
michel2323: I merged them.

src/Polar/kernels.jl
michel2323: Done

src/Polar/Constraints/line_flow.jl
```diff
@@ -30,13 +30,12 @@ function flow_constraints_grad!(polar::PolarForm, cons_grad, buffer, weights)
     fill!(cons_grad, 0)
     adj_vmag = @view cons_grad[1:nbus]
     adj_vang = @view cons_grad[nbus+1:2*nbus]
-    ev = adj_branch_flow_kernel!(polar.device)(weights, buffer.vmag, adj_vmag,
+    adj_branch_flow!(weights, buffer.vmag, adj_vmag,
```
frapac, line 33: How about renaming this function `adjoint_branch_flow!`, for clarity?

src/Polar/derivatives.jl
```diff
@@ -181,28 +184,24 @@ function AutoDiff.adj_hessian_prod!(
     nmap = length(H.map)
 
     # Init seed
-    for i in 1:nmap
+    @inbounds for i in 1:nmap
```
frapac, line 187: How fast it is on the GPU? Do you think it would be worth rewriting this for loop as a kernel?

test/Polar/hessian.jl
```diff
@@ -12,17 +14,22 @@ import ExaPF: PowerSystem, AutoDiff
...
-# Warning: currently works only on CPU, as depends on
-# explicit evaluation of Hessian, using MATPOWER expressions
 @testset "Compute reduced Hessian on CPU" begin
     @testset "Case $case" for case in ["case9.m", "case30.m"]
+        if has_cuda_gpu()
+            ITERATORS = zip([CPU(), CUDADevice()], [Vector, CuVector])
+        else
+            ITERATORS = zip([CPU()], [Vector])
+        end
+        @testset "Device $device" for (device, T) in ITERATORS
```
frapac, line 24: Could be indent the lines inside `@testset`?

test/Polar/hessian.jl
```diff
@@ -12,17 +14,22 @@ import ExaPF: PowerSystem, AutoDiff
...
         ##################################################
         # Initialization
         ##################################################
         datafile = joinpath(dirname(@__FILE__), "..", "..", "data", case)
         tolerance = 1e-8
         pf = PS.PowerNetwork(datafile)
-        polar = PolarForm(pf, CPU())
+        polar = PolarForm(pf, device)
+        cpu_polar = PolarForm(pf, CPU())
```
frapac, line 32: Oh, interesting! Why do we need a second `polar` on the CPU?

test/Polar/hessian.jl
```diff
@@ -296,19 +336,26 @@ const PS = PowerSystem
...
+        projp = Array(dev_projp)
         @test isapprox(projp, H_fd * tgt, rtol=1e-5)
 
         # Hessian w.r.t. Active-power generation
-        hess_pg = AutoDiff.Hessian(polar, ExaPF.active_power_constraints)
+        hess_pg = AutoDiff.Hessian(cpu_polar, ExaPF.active_power_constraints)
         μ = rand(1)
-        projp = AutoDiff.adj_hessian_prod!(polar, hess_pg, cache, μ, tgt)
-        ∂₂P = ExaPF.hessian(polar, ExaPF.active_power_constraints, cache, μ)
+        # TODO: No GPU implementation so far
```
frapac, line 350: ok!

test/Polar/hessian.jl
```diff
@@ -195,7 +223,8 @@ const PS = PowerSystem
         H_ffd = FiniteDiff.finite_difference_hessian(cost_x, [x; u])
 
         # Hessians of objective
-        ∇²f = ExaPF.hessian_cost(polar, cache)
+        # TODO: Active power on GPU
```
frapac, line 226: that should be easy, IMO

frapac (commented):
Nice job, I think we are almost there!

src/Polar/kernels.jl
```diff
@@ -1,6 +1,66 @@
 import KernelAbstractions: @index
 # Implement kernels for polar formulation
 
+"""
```
frapac, line 4: This function should be removed.

src/Polar/kernels.jl
```diff
@@ -1,6 +1,66 @@
...
+    get_power_injection(fr, v_m, v_a, ybus_re, ybus_im)
+
+Computes the power injection at node `fr`.
+In polar form, the power injection at node `i` satisfies
+```math
+p_{i} = \\sum_{j} v_{i} v_{j} (g_{ij} \\cos(\\theta_i - \\theta_j) + b_{ij} \\sin(\\theta_i - \\theta_j))
+```
+"""
+function get_power_injection(fr, v_m, v_a, ybus_re, ybus_im)
```
frapac, line 36: Idem

src/Polar/kernels.jl
```diff
@@ -1,6 +1,66 @@
...
+"""
+    get_react_injection(fr, v_m, v_a, ybus_re, ybus_im)
+
+Computes the reactive power injection at node `fr`.
+In polar form, the power injection at node `i` satisfies
+```math
+q_{i} = \\sum_{j} v_{i} v_{j} (g_{ij} \\sin(\\theta_i - \\theta_j) - b_{ij} \\cos(\\theta_i - \\theta_j))
+```
+"""
+function get_react_injection(fr::Int, v_m, v_a, ybus_re::Spmat{VI,VT}, ybus_im::Spmat{VI,VT}) where {VT <: AbstractVector, VI<:AbstractVector}
```
frapac, line 55: Idem

src/Polar/kernels.jl
```diff
@@ -125,15 +185,11 @@ KA.@kernel function adj_residual_edge_kernel!(
     adj_pinj[fr] -= adj_F[i]
 end
 
-KA.@kernel function adj_residual_node_kernel!(F, adj_F, vm, adj_vm, va, adj_va,
+KA.@kernel function cpu_adj_node_kernel!(F, adj_F, vm, adj_vm, va, adj_va,
```
frapac, line 188: I don't understand why we need two kernels for the CPU and the GPU here?

src/Polar/kernels.jl
```diff
@@ -560,60 +669,100 @@ KA.@kernel function adj_reactive_power_edge_kernel!(
     end
 end
 
+# Refresh active power (needed to evaluate objective)
+function adj_reactive_power!(
```
frapac, line 673: Rename as `adjoint_reactive_power!` ?

src/Polar/kernels.jl
```diff
@@ -144,31 +200,96 @@ KA.@kernel function adj_residual_node_kernel!(F, adj_F, vm, adj_vm, va, adj_va,
...
+    ev = kernel_node!(adj_vm, adj_va,
+            ybus_re.colptr, ybus_re.rowval,
+            edge_vm_from, edge_va_from, 
+            tspedge_vm_to.nzVal, tspedge_vm_to.rowInd, tspedge_vm_to.colInd,
+            tspedge_va_to.nzVal, tspedge_va_to.rowInd, tspedge_va_to.colInd,
+            ndrange=nvbus)
+    wait(ev)
+end
+
+function adj_residual_polar!(
```
frapac, line 270: Rename as `adjoint_residual_polar!`?

src/Polar/derivatives.jl
michel2323: The problem is that this has to be done on the CPU. I think let's take it in and get a general idea of the performance. There are other places in the code that may need work. I have no idea what the performance bottlenecks will be on larger cases.

michel2323: Oh the hanging from yesterday was my fault because I used `CPU()` somewhere where it should've been `CUDADevice()`.

src/Polar/kernels.jl
michel2323: Done.

src/Polar/kernels.jl
michel2323: Done.

src/Polar/kernels.jl
michel2323: Idem

src/Polar/kernels.jl
michel2323: This will be taken care off in a future naming PR.

src/Polar/kernels.jl
michel2323: Idem.

src/Polar/kernels.jl
michel2323: For the CPU kernel both edge metrices are in CSC. For the GPU kernel the transposed edge matrix in COO. You see it in the second loop of the GPU kernel which goes over all non-zeros.

test/Polar/hessian.jl
michel2323: Since we call the matpower hessians which can only be computed with a polar model on a CPU architecture. Maybe that's wrong.

src/Polar/Constraints/line_flow.jl
michel2323: We agreed that there will be a future naming PR.

test/Polar/hessian.jl
michel2323: Alright! Done.

src/Polar/kernels.jl
frapac: I would maybe add a comment to explain that, as this is non-trivial. What do you think?

test/Polar/hessian.jl
frapac: I see. I don't know what's best: either we compute Matpower Hessians on the CPU, or we don't compute them at all when we are testing on the GPU.

src/Polar/kernels.jl
michel2323: I commented a few functions including these, but then ran out of steam.

test/Polar/hessian.jl
michel2323: I think in principle it's not bad to instantiate both. This makes also sure that there are no side effects.

PR #115, opened 2021-03-03
Cleaning / Part I

frapac:
* remove deprecated getters in Polar
* clean Autodiff
* clean structure of `src/Polar/`
* move definition of MATPOWER's Jacobian and Hessian in `PowerSystem`


michel2323 (approved):

src/Polar/Constraints/active_power.jl
```diff
@@ -75,7 +75,7 @@ function jacobian(polar::PolarForm, cons::typeof(active_power_constraints), buff
     # Use MATPOWER to derive expression of Hessian
     # Use the fact that q_g = q_inj + q_load
     V = buffer.vmag .* exp.(im .* buffer.vang)
-    dSbus_dVm, dSbus_dVa = _matpower_residual_jacobian(V, polar.network.Ybus)
+    dSbus_dVm, dSbus_dVa = PS.matpower_residual_jacobian(V, polar.network.Ybus)
```
dmaldona, line 78: Great!

src/Polar/Constraints/line_flow.jl
```diff
@@ -74,32 +74,25 @@ function adjoint!(
     wait(ev)
 end
 
-function matpower_jacobian(polar::PolarForm, ::State, ::typeof(flow_constraints), V)
+function matpower_jacobian(polar::PolarForm, X::Union{State,Control}, ::typeof(flow_constraints), V)
```
dmaldona, line 77: @frapac Does this mean that this function definition is for both State and Control?

X::Union{State,Control}

src/Polar/Constraints/power_balance.jl
```diff
@@ -75,35 +75,42 @@ function adjoint!(
...
+        j22 = imag(dSbus_dVm[pq, pq])
+        return [j11 j12; j21 j22]
+    elseif isa(X, Control)
+        j11 = real(dSbus_dVm[[pv; pq], [ref; pv]])
+        j12 = sparse(I, npv + npq, npv)
+        j21 = imag(dSbus_dVm[pq, [ref; pv]])
+        j22 = spzeros(npq, npv)
+        return [j11 -j12; j21 j22]
+    end
+end
```
dmaldona, line 101: @frapac I actually prefer this version. Cool!

https://medium.com/@lovelydog/4-notes-modules-should-be-deep-ba5671c4288c

src/Polar/powerflow.jl
```diff
@@ -0,0 +1,122 @@
+
+function powerflow(
```
dmaldona, line 2: @frapac when we do include("powerflow.jl") and this function is included into the module. How can the user know where to look for the function?

src/Polar/Constraints/line_flow.jl
frapac: Indeed: we could use this function either with `State` and `Control`. The reason why I am doing that, instead of coding two functions
```julia
matpower_jacobian(..., X::State, ...)
matpower_jacobian(..., X::Control, ...)
```
is to avoid duplicate code, as the implementations of the two functions would look almost similar. 
Performance wise, the compiler is clever enough to handle the switch efficiently (I have a reference about that, let me share it with you)

dmaldona (approved):
Good job! Everything looks great. I like that you're condensing functions instead of further partitioning functionality.

src/Polar/powerflow.jl
frapac: Good question. We include `powerflow` here: https://github.com/exanauts/ExaPF.jl/blob/fp/cleaning/src/Polar/polar.jl#L32
But I think this is not satisfactory, as it is a bit obfuscated inside the code of `src/Polar/polar.jl`...

frapac:
> I like that you're condensing functions instead of further partitioning functionality.

I think your former remark was legitimate, and the code was starting to look like pasta. :spaghetti:

PR #116, opened 2021-03-03
Cleaning / Part II:  move Parsers and Indexes into PowerSystem

frapac:
Independent from #115 

- Move `Parsers` and `Indexes` inside `PowerSystem` submodule
- now `PowerSystem` is standalone, and independent from the other modules in ExaPF


michel2323:
Maybe we should merge this. @dmaldona you have any objections?

PR #117, opened 2021-03-05
Create system image

michel2323:
This creates a system image via `make image`. The image will be the file `exapf.so` and it will only be created if a julia source file in `src` is newer than the image file.

Big caveat: The MOI test breaks the image creating. So, this test will have to be manually uncommented. Unfortunately, I have not found a way to detect from within the tests whether they are called from `PackageCompiler`. I tried `stacktrace()` etc.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+data/* linguist-documentation
+scripts/* linguist-documentation
+docs/* linguist-documentation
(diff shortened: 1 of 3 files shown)


frapac (approved):
I would maybe move `createsysimage.jl` into a new directory `deps/`

michel2323:
> I would maybe move `createsysimage.jl` into a new directory `deps/`

Fixed.

michel2323:
Added a warning telling the user to uncomment the MOI test.

PR #121, opened 2021-03-10
MOI: load Hessian only if available

frapac:


michel2323 (approved):
This looks like a no-brainer.

PR #122, opened 2021-03-10
Tolerances needed for tests to pass on Summit

michel2323:
These are the tolerances needed for the tests to pass on Summit. I am not sure this is reasonable.


frapac:
Commit included in #118

PR #127, opened 2021-03-15
CI update

michel2323:
The CI is testing Julia 1.6 and the Ubuntu version is set to `20.04` to not one day get a bad surprise.


frapac (approved):
Looks good to me!

PR #130, opened 2021-03-17
Fix bounds type error on CUDADevice()

michel2323:


frapac (approved):
Nice catch!

michel2323:
This has to be resolved more consistently. Cancelling merge for now.

PR #131, opened 2021-03-18
Non-allocating kernels for adjoints

frapac:
This PR removes most of the allocations that occurred in the kernels, when: 
- we compute explicitly a transpose matrix each time we called the adjoints of `reactive_power` or `power_balance`
- we allocate 4 intermediate caches each time we called the adjoints of `reactive_power`, `power_balance`, `line_flow`

We introduce a new `PullbackMemory` object to cache the intermediate caches needed in the computation:
- now, the functions `adjoint!` and `jtprod!` takes as input a `PullbackMemory` object
- the Hessian has been modified and has its own `PullbackMemory`, of `Dual` type

The pullback memory object is greatly inspired from [ChainRulesCore](https://juliadiff.org/ChainRulesCore.jl/stable/design/changing_the_primal.html#The-Journey-to-rrule)
The benchmarks are promising, and we are now almost on par with Matpower's Hessian. For `case300.m`, on the CPU, we get

- Original Matpower's Hessian
```
  Hessprod          194.124 μs (34 allocations: 168.84 KiB)
```
- First implementation of Hessian with AutoDiff: 
```
  Hessprod          369.569 μs (98 allocations: 222.25 KiB)
```
- Removing the allocation in the kernels (this PR)
```
   Hessprod          319.977 μs (51 allocations: 88.39 KiB)
```
- Combining with PR #128 
```
Hessprod          299.846 μs (32 allocations: 67.56 KiB)

```

### DIFF
--- a/src/Evaluators/reduced_evaluator.jl
+++ b/src/Evaluators/reduced_evaluator.jl
@@ -42,8 +42,8 @@ mutable struct ReducedSpaceEvaluator{T, VI, VT, MT, Jacx, Jacu, JacCons, Hess} <
     buffer::PolarNetworkState{VI, VT}
     # AutoDiff
     state_jacobian::FullSpaceJacobian{Jacx, Jacu}
-    obj_stack::AutoDiff.PullbackMemory{typeof(active_power_constraints), AdjointStackObjective{VT}, Nothing}
-    cons_stacks::Vector{AutoDiff.PullbackMemory} # / constraints
+    obj_stack::AutoDiff.TapeMemory{typeof(active_power_constraints), AdjointStackObjective{VT}, Nothing}
+    cons_stacks::Vector{AutoDiff.TapeMemory} # / constraints
     constraint_jacobians::JacCons
     hessians::Hess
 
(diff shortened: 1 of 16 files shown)


src/Evaluators/reduced_evaluator.jl
```diff
@@ -82,12 +82,12 @@ function ReducedSpaceEvaluator(
...
-    obj_ad = AdjointStackObjective(model)
+    obj_ad = pullback_objective(model)
     state_ad = FullSpaceJacobian(model, power_balance)
     VT = typeof(g_min)
-    cons_ad = AdjointPolar{VT}[]
+    cons_ad = AutoDiff.PullbackMemory[]
     for cons in constraints
-        push!(cons_ad, AdjointPolar(model))
+        push!(cons_ad, AutoDiff.PullbackMemory(model, cons, VT))
     end
```
michel2323, line 91: In AD literature this is usually called stack or tape. But pullback seems to be the new Julia slang. Below where what is now called the stack is defined I made another comment.

src/autodiff.jl
```diff
@@ -89,22 +87,32 @@ Creates an object for computing Hessian adjoint tangent projections.
...
     map::VI
     varx::SubT
     t1svarx::SubD
+    buffer::Buff
 end
 
+# Cache for adjoint
+# Largely inspired from ChainRulesCore.jl:
+# https://juliadiff.org/ChainRulesCore.jl/stable/design/changing_the_primal.html#The-Journey-to-rrule
+struct PullbackMemory{F, S, I}
```
michel2323, line 108: This makes total sense now. You essentially came up with the stack that is needed to save the entire state of the computation. That is: function, value, partial. So what is called stack here should probably be a partial. And `PullbackMemory` is what is called tape or stack in AD.

michel2323 (approved):

src/Evaluators/reduced_evaluator.jl
frapac: I have renamed `PullbackMemory` as `TapeMemory`

PR #133, opened 2021-03-22
[MOI] clean Hessian in MathOptInterface wrapper

frapac:
* Avoid creating a new matrix each time a Hessian or a Jacobian callback is being called


frapac:
The previous version of this PR introduced a performance regression. This has been addressed in this new version.

michel2323 (approved):

Issue #138, opened 2021-03-29
Unable to run code in the Quick Start


lcw:
In reviewing your JuliaCon submission, https://github.com/JuliaCon/proceedings-review/issues/72, I have been exercising you code.  I have been unable to run the code described in the quick start, both on the master branch and on the develop branch.

First I started making changes like the following
```diff
diff --git a/docs/src/quickstart.md b/docs/src/quickstart.md
index a9fc2c7..491fe75 100644
--- a/docs/src/quickstart.md
+++ b/docs/src/quickstart.md
@@ -69,6 +69,7 @@ in a few lines of code.
 We first instantiate a `PolarForm` object to adopt a polar formulation
 as a model:
 ```julia-repl
+julia> using KernelAbstractions
 julia> polar = PolarForm(pf, CPU())
 
 ```
@@ -90,7 +91,7 @@ Hence, the algorithm requires the following elements:
 
 that translate to the Julia code:
 ```julia-repl
-julia> physical_state = get(polar, PhysicalState())
+julia> physical_state = get(polar, ExaPF.PhysicalState())
 julia> jx = ExaPF.init_ad_factory(polar, physical_state)
 julia> linear_solver = DirectSolver()
```
but I ran into the issue that `ERROR: UndefVarError: init_ad_factory not defined`.  I am stuck here.

p.s. it looks like the nested code blocks are messing up the formatting, not sure how to fix that.

frapac:
Dear Lucas,
thanks for pointing out this issue. 
we are sorry for the inconvenience. ExaPF has been refactored a lot for the last three months, and we haven't proceed to any release since December (hence the slightly deprecated doc). 

We should be ready to do a release `v0.5.0` in the next few days. At least, that will ensure that everything is up to date on master. The documentation has also been updated recently: #134 

I was not aware that the nested code blocks are messing up the formatting. I will investigate that.

lcw:
> I was not aware that the nested code blocks are messing up the formatting. I will investigate that.

This was only for my diff above.

frapac:
Closing as this issue was solved in #140

PR #140, opened 2021-03-30
Update quickstart guide

frapac:
The goal of this PR is to keep the quickstart guide up to date, which was not the case before. This was a major issue, as illustrated in #138 

* add instance case1354_pegase in data/


docs/src/quickstart.md
```diff
@@ -3,20 +3,29 @@
 This page introduces the first steps to set up `ExaPF.jl`.
 We show how to load a power network instance and how to solve
 the power flow equations both on the CPU and on the GPU.
+The full script is implemented in [this file](https://github.com/exanauts/ExaPF.jl/tree/master/test/quickstart.jl)
 
-### How to load a MATPOWER instance?
-We start by importing into `ExaPF` an instance specified in the MATPOWER format.
+We start by importing CUDA and KernelAbstractions:
+```julia-repl
+julia> using CUDA
```
michel2323, line 10: I think it would be better to remote `julia>` as this allows a user to copy paste the commands, like here in [CUDA.jl](https://juliagpu.github.io/CUDA.jl/stable/)

docs/src/quickstart.md
```diff
@@ -85,16 +94,25 @@ The algorithm solves at each step the linear equation:
...
 
 ```
+
+We build a Jacobian object storing all structures needed by
+the AutoDiff backend:
+```julia-repl
+julia> jx = AutoDiff.Jacobian(polar, ExaPF.power_balance, State())
+A AutoDiff Jacobian for power_balance (w.r.t. State())
+Number of Jacobian colors: 25
+```
```
michel2323, line 114: I see you use `julia-repl` here. I think we should separate the output from the code though for the same reasons than above.

docs/src/quickstart.md
```diff
@@ -198,7 +227,7 @@ julia> pf_algo = NewtonRaphson(; verbose=1, tol=1e-7)
 Giving:
 ```julia-repl
 julia> convergence = ExaPF.powerflow(polar_gpu, jx_gpu, physical_state_gpu, pf_algo;
-                                     linear_solver=linear_solver)
+                                     solver=linear_solver)
 Iteration 0. Residual norm: 26.6667.
```
michel2323, line 231: Same with the output. I'd split it up.

michel2323 (approved):

PR #147, opened 2021-04-02
Improve performance of reduced Hessian on GPU

frapac:
- in reverse over forward pass, call proper kernel to compute the Hessian of the objective
- transpose matrix implicitly in the adjoint of the residual kernel (instead of using views) 
- replace `dotview` by `copyto!`, leading to better performance on the GPU

### DIFF
--- a/src/Evaluators/reduced_evaluator.jl
+++ b/src/Evaluators/reduced_evaluator.jl
@@ -78,7 +78,7 @@ mutable struct ReducedSpaceEvaluator{T, VI, VT, MT, Jacx, Jacu, JacCons, Hess} <
     buffer::PolarNetworkState{VI, VT}
     # AutoDiff
     state_jacobian::FullSpaceJacobian{Jacx, Jacu}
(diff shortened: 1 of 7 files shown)


src/Evaluators/reduced_evaluator.jl
```diff
@@ -512,8 +512,8 @@ function hessprod!(nlp::ReducedSpaceEvaluator, hessvec, u, w)
     _second_order_adjoint_z!(nlp, z, w)
 
     # Init tangent
-    tgt[1:nx] .= z
-    tgt[1+nx:nx+nu] .= w
+    copyto!(tgt, 1, z, 1, nx)    # tgt[1:nx] .= z
```
michel2323, line 515: To me, it does not have to be explained in a comment anymore.

michel2323 (approved):

src/Evaluators/reduced_evaluator.jl
frapac: I will remove the comment then

frapac: Comments have been removed.

PR #149, opened 2021-04-12
Minor documentation updates

michel2323:
@frapac There are some missing docstrings for functions/objects that are in the documentation. Those are:
* `update!`
* `AbstractNetworkBuffer`

and in the quickstart there's a reference to `AbstractPhysicalCache` which I am not sure still exists.


frapac (approved):
Looks good to me!

PR #151, opened 2021-04-13
Update to KA 0.6

michel2323:
Update to KernelAbstractions 0.6.

* No dependency on `CUDAKernels.jl`
* Removed any occurence of `CUDADevice`
* ExaPF now distinguishes only between `KA.CPU()` and `KA.GPU()` device. Of course this is only valid if the GPU is always a CUDA device for now. This will be extended when moving to AMD GPUs.

### DIFF
--- a/src/LinearSolvers/LinearSolvers.jl
+++ b/src/LinearSolvers/LinearSolvers.jl
@@ -44,8 +44,7 @@ abstract type AbstractIterativeLinearSolver <: AbstractLinearSolver end
 """
     list_solvers(::KernelAbstractions.Device)
 
-List linear solvers available on current device. Currently,
-only `CPU()` and `CUDADevice()` are supported.
+List linear solvers available on current device.
 
 """
 function list_solvers end
@@ -137,8 +136,8 @@ function ldiv!(::DirectSolver{Nothing},
 end
 get_transpose(::DirectSolver, M::CUSPARSE.CuSparseMatrixCSR) = CUSPARSE.CuSparseMatrixCSC(M)
 
-function update!(solver::AbstractIterativeLinearSolver, J)
-    update(J, solver.precond)
+function update_preconditioner!(solver::AbstractIterativeLinearSolver, J, device)
+    update(solver.precond, J, device)
 end
 
 """
@@ -270,10 +269,9 @@ List all linear solvers available solving the power flow on the CPU.
 list_solvers(::KA.CPU) = [RefBICGSTAB, RefGMRES, DQGMRES, BICGSTAB, EigenBICGSTAB, DirectSolver, KrylovBICGSTAB]
 
 """
-    list_solvers(::KA.CUDADevice)
+    list_solvers(::KA.GPU)
 
 List all linear solvers available solving the power flow on an NVIDIA GPU.
 """
-list_solvers(::KA.CUDADevice) = [BICGSTAB, DQGMRES, EigenBICGSTAB, DirectSolver, KrylovBICGSTAB]
-
+list_solvers(::KA.GPU) = [BICGSTAB, DQGMRES, EigenBICGSTAB, DirectSolver, KrylovBICGSTAB]
 end
--- a/src/LinearSolvers/preconditioners.jl
+++ b/src/LinearSolvers/preconditioners.jl
@@ -42,22 +42,22 @@ struct BlockJacobiPreconditioner{AT,GAT,VI,GVI,MT,GMT,MI,GMI,SMT} <: AbstractPre
     P::SMT
     id::Union{GMT,MT}
     function BlockJacobiPreconditioner(J, npart, device=CPU()) where {}
-        if device == CPU()
+        if isa(device, CPU)
             AT  = Array{Float64,3}
             GAT = Nothing
             VI  = Vector{Int64}
             GVI = Nothing
-            MT = Matrix{Float64}
+            MT  = Matrix{Float64}
             GMT = Nothing
             MI  = Matrix{Int64}
-            GMI  = Nothing
+            GMI = Nothing
             SMT = SparseMatrixCSC{Float64,Int64}
-        elseif device == CUDADevice()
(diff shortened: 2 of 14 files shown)


frapac (approved):
Looks good to me!

src/LinearSolvers/LinearSolvers.jl
```diff
@@ -137,8 +136,8 @@ function ldiv!(::DirectSolver{Nothing},
 end
 get_transpose(::DirectSolver, M::CUSPARSE.CuSparseMatrixCSR) = CUSPARSE.CuSparseMatrixCSC(M)
 
-function update!(solver::AbstractIterativeLinearSolver, J)
-    update(J, solver.precond)
+function precond_update!(solver::AbstractIterativeLinearSolver, J, device)
```
frapac, line 139: I am not sure about the naming of this function. I will keep `update!` (or at least rename it `update_preconditioner!` for consistency). 
Also: do we really need to pass the device here? I guess the device could be inferred by dispatching on the type of `J`

src/LinearSolvers/preconditioners.jl
```diff
@@ -288,16 +288,16 @@ Update the preconditioner `p` from the sparse Jacobian `J` in CSC format for the
 Note that this implements the same algorithm as for the GPU and becomes very slow on CPU with growing number of blocks.
 
 """
-function update(J::SparseMatrixCSC, p)
-    kernel! = update_cpu_kernel!(CPU())
+function update(p, J::SparseMatrixCSC, device)
```
frapac, line 291: The ordering of the argument is better, indeed. I would move one step further, and suggest the signature:
```julia
function update!(p, J::SparseMatrixCSC)
```
and remove the device in the argument

src/LinearSolvers/preconditioners.jl
```diff
@@ -288,16 +288,16 @@ Update the preconditioner `p` from the sparse Jacobian `J` in CSC format for the
...
+    kernel! = update_cpu_kernel!(device)
     p.P.nzval .= 0.0
-    ev = kernel!(J.colptr, J.rowval, J.nzval, p, p.lpartitions, ndrange=p.nblocks, dependencies=Event(CPU()))
+    ev = kernel!(J.colptr, J.rowval, J.nzval, p, p.lpartitions, ndrange=p.nblocks, dependencies=Event(device))
     wait(ev)
     return p.P
 end
-function update(J::Transpose{T, SparseMatrixCSC{T, I}}, p) where {T, I}
+function update(p, J::Transpose{T, SparseMatrixCSC{T, I}}) where {T, I}
```
frapac, line 298: Not sure this function is tested in the tests.

src/architectures.jl
```diff
@@ -2,7 +2,7 @@
 abstract type AbstractArchitecture end
 
 array_type(::KA.CPU) = Array
-array_type(::KA.CUDADevice) = CUDA.CuArray
+array_type(::KA.GPU) = CUDA.CuArray
```
frapac, line 5: not sure we want to fallback on `CuArray` here?

src/LinearSolvers/preconditioners.jl
```diff
@@ -42,22 +42,22 @@ struct BlockJacobiPreconditioner{AT,GAT,VI,GVI,MT,GMT,MI,GMI,SMT} <: AbstractPre
...
             GVI = Nothing
-            MT = Matrix{Float64}
+            MT  = Matrix{Float64}
             GMT = Nothing
             MI  = Matrix{Int64}
-            GMI  = Nothing
+            GMI = Nothing
             SMT = SparseMatrixCSC{Float64,Int64}
-        elseif device == CUDADevice()
+        elseif isa(device, GPU)
```
frapac, line 55: Do we want to use `CuArray` here?

frapac (changes requested):
That's a nice rescope!
However, I think the PR is not ready to merge on `develop`, yet. Apart of the few changes suggested in the review, I am wondering whether we should encapsulate properly all codes related to CUDA to ease the integration with AMDGPU.jl.

src/LinearSolvers/LinearSolvers.jl
michel2323: I found the `update` functions confusing. In addition, what is updated is not even in the signature (linear solver vs preconditioner). I would vote for `update_preconditioner!` then.

src/LinearSolvers/preconditioners.jl
michel2323: I think we should use device consistently throughout the code. I think it shouldn't be decided based on the datatype. There is not safe map from datatype to device type. Oceananigans uses `arch` throughout the code. I like `device` better.

src/LinearSolvers/preconditioners.jl
michel2323: I don't even know what it is used for :-) . Shouldn't we add a test then?

src/architectures.jl
michel2323: It is for now everywhere in the code. As I described above, this should work for CUDA and CPU only. I think we should wait until we move something actually to AMDGPU.jl.

src/LinearSolvers/preconditioners.jl
michel2323: Yes. It's used throughout the code. This is the place where `Preferences.jl` would come in, but I don't want to add that now. I want to add that when at least one kernel has a test on AMD GPUs.

michel2323:
The changes to `AMDGPU.jl` requires the addition of something similar to `Preferences.jl`. I think we should use KA 0.6 before we're adding AMDGPU support. The only places in the code that will have to be adapted are the type definitions and various forks in the code based on the device,  basically anywhere `isa(x, GPU)` is used. I just would not like to force that on `develop` just yet.

src/LinearSolvers/LinearSolvers.jl
frapac: sounds good to me!

PR #155, opened 2021-04-21
Allow to run GPU kernels in batch

frapac:
This PR modifies the kernels so they can process batch of inputs, instead of a single input as previously. The AutoDiff has been updated accordingly. 

This new feature allows to 
- run the powerflow in batch (with a new `batch_powerflow` function) 
- compute the reduced Hessian in batch mode


michel2323:
This looks good to me. As we discussed it of course duplicates a bit of code, but I think we can resolve that at some later point when we have a project with these batched powerflows.

michel2323 (approved):

PR #156, opened 2021-04-22
Refactoring of tests directory

frapac:
Large refactoring of tests directory, to ease the integration with AMDGPU.jl 

* now, the tests for `PolarForm` and the `Evaluators` are implemented inside two dedicated modules: `TestPolarFormulation` and `TestEvaluators`. All tests implemented the two modules are independent of the architecture (CPU, CUDADevice ...) and the input file (`case9.m`, `case30.m`) 
* all tests are now encapsulated in proper functions. We don't need to run `include("test/runtests.jl")` everytime we want to test a single functionality, and we could select the tests we want to launch directly in the REPL.
* to launch all the tests for `PolarForm`: 
```julia
TestPolarFormulation.runtests("data/case9.m", CPU(), Array) # on the CPU
TestPolarFormulation.runtests("data/case9.m", CUDADevice(), CuArray) # on the GPU
TestPolarFormulation.runtests("data/case57.m", CPU(), Array) # on a different case
```
* same principle for `Evaluator`: 
```julia
TestEvaluators.runtests("data/case9.m", CPU(), Array) # on the CPU
TestEvaluators.runtests("data/case9.m", CUDADevice(), CuArray) # on the GPU
TestEvaluators.runtests("data/case57.m", CPU(), Array) # on a different case
```
It remains to rewrite `test/runtests.jl`, as I think [the current file](https://github.com/exanauts/ExaPF.jl/blob/fp/clean_tests/test/runtests.jl#L24-L55) is not readable. Any feedback welcome on that!


michel2323 (approved):

michel2323:
Nice!

PR #161, opened 2021-04-29
Remove IterativeSolvers from ExaPF

frapac:
I don't think we need IterativeSolvers anymore, as we are now using Krylov.jl, which comes at the advantage of being compatible  with the GPU.


dmaldona:
Good idea!

michel2323 (approved):

Issue #28, opened 2020-08-23
Multiple generators per bus


dmaldona:
When we solve the power flow we usually have one equation per bus - we merge all the power injections in a bus. 

In the OPF we have to keep track of each generator power injection to include in the cost function.

Or current formulation is unable to deal with this case. This affects the pegase and rte cases.

frapac:
Supported with the new formulation

PR #168, opened 2021-05-10
Multiple generators, part I:  Implement aggregation of multiple generators

frapac:
Allow to support cases with multiple generators in ExaPF. 

If a bus has multiple generators, we aggregate the data and the costs coefficients corresponding to the generators, so we get only one generator at each PV bus.

See #28

### DIFF
--- a/src/PowerSystem/power_network.jl
+++ b/src/PowerSystem/power_network.jl
@@ -73,7 +73,7 @@ struct PowerNetwork <: AbstractPowerSystem
 
         # COSTS
         if isnothing(cost_coefficients)
-            println("[PS] Cost function not specified in dataset. Fallback to default coefficients.")
+            @warn("[PS] Cost function not specified in dataset. Fallback to default coefficients.")
             # if not specified, costs is set by default to a # quadratic polynomial
             costs = zeros(ngen, 7)
             costs[:, 1] .= 2.0 # polynomial model
@@ -109,7 +109,7 @@ struct PowerNetwork <: AbstractPowerSystem
         ref_id = bus[ref, 1]
         @assert bus[ref, 2] == [REF_BUS_TYPE]
         if !(ref_id[1] in gen[:, 1])
-            println("[PS] No generator affected to slack node.")
+            error("[PS] No generator attached to slack node.")
         end
 
         sbus, sload = assembleSbus(gen, bus, SBASE, bus_id_to_indexes)
--- a/src/PowerSystem/power_network.jl
+++ b/src/PowerSystem/power_network.jl
@@ -40,7 +40,7 @@ struct PowerNetwork <: AbstractPowerSystem
     sbus::Vector{Complex{Float64}}
     sload::Vector{Complex{Float64}}
 
-    function PowerNetwork(data::Dict{String, Array}; remove_lines=Int[])
+    function PowerNetwork(data::Dict{String, Array}; remove_lines=Int[], multi_generators=:aggregate)
         # Parsed data indexes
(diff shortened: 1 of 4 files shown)


src/PowerSystem/power_network.jl
```diff
@@ -40,7 +40,7 @@ struct PowerNetwork <: AbstractPowerSystem
     sbus::Vector{Complex{Float64}}
     sload::Vector{Complex{Float64}}
 
-    function PowerNetwork(data::Dict{String, Array}; remove_lines=Int[])
+    function PowerNetwork(data::Dict{String, Array}; remove_lines=Int[], multi_generators=:aggregate)
```
dmaldona, line 43: @frapac aggregate in this case will be what we have been doing until now. Right?

frapac: Exactly. We aggregate the generators to get one generator per PV/slack node, at the end. It's mostly a Julia port of your matlab script, I would say.

dmaldona: @frapac got it. thanks!!

src/PowerSystem/power_network.jl
```diff
@@ -50,17 +50,45 @@ struct PowerNetwork <: AbstractPowerSystem
...
+        # LINES
         # Remove specified lines
         lines = get_active_branches(lines, remove_lines)
 
         # size of the system
         nbus = size(bus, 1)
         ngen = size(gen, 1)
 
+        # COSTS
+        if isnothing(cost_coefficients)
```
dmaldona, line 75: @frapac In the message. Could you specify the generator number which does not have the coefficients? Also, can we show warnings instead of printing it directly to STDOUT?

src/PowerSystem/power_network.jl
```diff
@@ -77,6 +105,12 @@ struct PowerNetwork <: AbstractPowerSystem
 
         # bus type indexing
         ref, pv, pq, bustype = bustypeindex(bus, gen, bus_id_to_indexes)
+        # check consistency
+        ref_id = bus[ref, 1]
+        @assert bus[ref, 2] == [REF_BUS_TYPE]
+        if !(ref_id[1] in gen[:, 1])
+            println("[PS] No generator affected to slack node.")
```
dmaldona, line 112: @frapac here I would say "attached" in place of "affected". 

Perhaps we can throw an error, I am not sure what behavior we should have in this case.

src/PowerSystem/power_network.jl
```diff
@@ -85,9 +119,9 @@ struct PowerNetwork <: AbstractPowerSystem
     end
 end
 
-function PowerNetwork(datafile::String)
+function PowerNetwork(datafile::String; options...)
     data = import_dataset(datafile)
```
dmaldona, line 123: @frapac ah, so elegant!

src/PowerSystem/power_network.jl
dmaldona: Below I see that you use the macro "@warn"

dmaldona (approved):
Great work! I just added some minor comments.

src/PowerSystem/power_network.jl
frapac: Pythonic style ;-)

src/PowerSystem/power_network.jl
frapac: Good catch. "attached" is better than "affected". I will replace the print by throwing an error there.

src/PowerSystem/power_network.jl
frapac: Again, good catch! I will ensure that the warning are printed in STDERR. 
Just: if `cost_coefficient` is equal to `nothing`, that means that all costs are equal to 0. It happens when we are importing PSSE file, as we are not importing the attached file corresponding to the cost.

PR #173, opened 2021-06-01
Allow CUDA 2.x

michel2323:
Only CUDA 3.x is a bit harsh and lead to upstream compatibility issues (e.g. last release of MadNLP).


frapac (approved):
I would maybe a bit more restrictive. I remember that we should support CUDA.jl starting from something like `~2.3`

michel2323:
> I would maybe a bit more restrictive. I remember that we should support CUDA.jl starting from something like `~2.3`

Fixed.

PR #175, opened 2021-06-03
Integrate generic backend for deployment on AMD GPUs (WIP)

michel2323:
`architectures.jl` has been renamed to `backends.jl`. We now distinguish between 

`device`: The actual device a kernel runs on, 
`backend`: The backend API that is used.

In theory, OneAPI for example could support both CPUs and GPUs as a backend. The backend defines the API, whereas the device defines the hardware.

As is, this PR adds `AMDGPU.jl` as a dependence alike to `CUDA.jl`. `CUDAKernels.jl` and `ROCKernels.jl` are only loaded if a supported device is found. 

So far only the Eigen and Krylov.jl iterative solvers pass the test on ROCm.

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -12,6 +12,8 @@ on:
 
 jobs:
   test-github-cpuonly:
+    env:
+      JULIA_AMDGPU_DISABLE_ARTIFACTS: 1
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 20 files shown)


michel2323:
Converted back to draft as portability is not yet in the works for second-order optimization methods.

michel2323:
Done by https://github.com/exanauts/ExaPF.jl/pull/280

PR #176, opened 2021-06-03
Switch to Krylov.jl 0.7.3

frapac:
Krylov.jl adds a new feature that is of direct interest for us: inplace BICGSTAB. 
- Indeed, the BICGSTAB algorithm is using some intermediate arrays. When we were solving the powerflow equations, we had to allocate these auxiliary arrays each time we were calling `Krylov.bicgstab`. 
- This PR releases this burden, as we are using the new API of Krylov.jl to instantiate the intermediate arrays once and for all, at the beginning. Then, everything runs like a bliss

But ... we could not merge this PR as is. We need to discuss the API to instantiate a new `AbstractLinearSolver` instance inside ExaPF. Currently, we are calling
```julia
lin_solver = LinearSolvers.KrylovBICGSTAB(J, precond)
```
with `J` power flow Jacobian, and `precond` a preconditioner. 
I would suggest we use instead: 
```julia
lin_solver = LinearSolvers.KrylovBICGSTAB(J; precond=:block_jacobi, npartitions=100)
```
and then instantiate the preconditioner directly inside the constructor of `KrylovBICGSTAB` (and other linear solvers). 
What do you think?


dmaldona (commented):
Great PR!

michel2323 (approved):

PR #177, opened 2021-06-16
[REVIEW] Batch Hessian - Part I

frapac:
This PR implements the full batch Hessian operation in `src/Polar`. 

*tl;dr* 
In the optimization algorithm, we need to evaluate the Hessian for the objective, and for the constraints.
- Before that PR, the Hessians for the objective and the constraints were evaluated separately, leading to a lot of duplicate operations. The result was quite inefficient.
- Now, this PR uses a single kernel to evaluate all the Hessians together, using a single GPU kernel.

*Details*
- add a new structure `HessianLagrangian` to evaluate the reduced Hessian in the reduced space
- add three new functions `network_operations`, `cost_penalty_ramping_constraints` and `bus_injection`, with associated kernels
- rename objective function in `src/Polar` (from `active_power_generation` to `cost_production`)
- update `ProxALEvaluators` to use directly `cost_penalty_ramping_constraints` (so now ProxAL's objective is fully computed with AutoDiff)

### DIFF
--- a/src/Evaluators/proxal_evaluators.jl
+++ b/src/Evaluators/proxal_evaluators.jl
@@ -18,8 +18,10 @@ Evaluator wrapping a `ReducedSpaceEvaluator` for use inside the
 decomposition algorithm implemented in [ProxAL.jl](https://github.com/exanauts/ProxAL.jl).
 
 """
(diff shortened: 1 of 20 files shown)


michel2323 (approved):

frapac:
Closing this PR in favor of #179

PR #179, opened 2021-06-18
Integrate batch Hessian in `ReducedSpaceEvaluator` routine

frapac:
This PR allows to use the batch Hessian in any optimization routine. 

Further: 
- we can now evaluate the Jacobian on the GPU, in batch
- we use some metaprogramming to generate automatically the different batch routines in `src/Evaluators/reduced_evaluator.jl`
- we remove all the previous code that we used before to compute the reduced Hessian (inefficient on the GPU)
- add a `BridgeDeviceEvaluator` to wrap an Evaluator on the GPU and move the results back to the CPU


frapac:
Side note: this PR target the PR#178, to ease the reviewing process.

.github/workflows/action.yml
```diff
@@ -22,6 +22,7 @@ jobs:
       - uses: julia-actions/setup-julia@latest
         with:
           version: ${{ matrix.julia-version }}
+      - run: julia --project -e 'using Pkg; Pkg.add(url="https://github.com/exanauts/BlockPowerFlow.jl", rev="master")'
```
michel2323, line 25: We might want to use a consistent workflow with a `deps/deps.jl` script to pull in the required dependencies. Let me know whether I should push that to this branch.

Project.toml
```diff
@@ -19,7 +20,7 @@ SparseDiffTools = "47a9eef4-7e08-11e9-0b38-333d64bd3804"
 TimerOutputs = "a759f4b9-e2f1-59dc-863e-4aeb61b1ea8f"
 
 [compat]
-CUDA = "~2.3, 3.0"
+CUDA = "~2.3, ~3.2"
```
michel2323, line 23: Sorry, this creates a conflict with my push to develop where I forced `2.6` instead of `2.3`. `2.3` has issues on Summit.

src/Evaluators/bridge_evaluator.jl
```diff
@@ -0,0 +1,138 @@
+
+#=
+    BridgeEvaluator
+=#
```
michel2323, line 4: I am still not sure what the role of the `BridgeEvaluator` is. It uses a model on a device (e.g. `PolarForm`) and returns the output data of the evaluation functions on the selected device?

src/Evaluators/common.jl
```diff
@@ -112,7 +111,9 @@ function _inf_pr(nlp::AbstractNLPEvaluator, cons)
     return max(err_inf, err_sup)
 end
 
-# Scaler utils
+#=
+    SCALER
+=#
 abstract type AbstractScaler end
```
michel2323, line 117: `Inf` norm is called in here. It might be worthwhile implementing a kernel for that using `abs.(x)` and `reduce(max, x)`. Not sure though.

test/Evaluators/api.jl
```diff
@@ -87,14 +87,12 @@ function test_evaluator_callbacks(nlp, device, M; rtol=1e-6)
...
-            # Test Jacobian vector product
-            ExaPF.jprod!(nlp, v, u, jv)
-            @test isapprox(J * jv, v)
-        end
+        # J = ExaPF.jacobian(nlp, u)
+        # # Test transpose Jacobian vector product
+        # @test isapprox(jv, J' * v)
+        # # Test Jacobian vector product
+        # ExaPF.jprod!(nlp, v, u, jv)
+        # @test isapprox(J * jv, v)
```
michel2323, line 95: Why is this commented out?

src/Evaluators/reduced_evaluator.jl
```diff
@@ -435,141 +417,215 @@ end
...
+    H = nlp.hesslag
+    batch_adj_hessian_prod!(nlp.model, H.hess, hv, nlp.buffer, y, tgt)
+    ∂fₓ = hv[1:nx, :]
+    ∂fᵤ = hv[nx+1:nx+nu, :]
+    return ∂fₓ , ∂fᵤ
 end
 
 function hessprod!(nlp::ReducedSpaceEvaluator, hessvec, u, w)
-    @assert nlp.hessians != nothing
+    @assert nlp.hesslag != nothing
```
michel2323, line 441: `!isnothing()` I think is better suited.

benchmark/batch_hessian.jl
```diff
@@ -0,0 +1,87 @@
...
+    powerflow_solver = NewtonRaphson(tol=1e-10)
+
+    J = ExaPF.jacobian_sparsity(polar, ExaPF.power_balance, State())
+    if isa(device, CPU)
+        lufac = lu(J)
+        linear_solver = LS.DirectSolver(lufac)
+        if nbatch == 1
+            hessian_lagrangian = ExaPF.HessianLagrangian(polar, lufac, lufac')
+        else
+            hessian_lagrangian = ExaPF.BatchHessianLagrangian(polar, lufac, lufac', nbatch)
```
michel2323, line 29: Isn't `HessianLagrangian` the same as `BatchHessianLagragian` with `nbatch=1? Why do we have to make it a special case?

michel2323 (approved):

test/Evaluators/api.jl
frapac: Right, I should resolve that

benchmark/batch_hessian.jl
frapac: Good question. In fact, I think defining two separate structures (`BatchHessianLagrangian` and `HessianLagrangian`) is making sense, as 
1. it allows to use multiple dispatch to dispatch either using batch or single Hessian-vector products to compute the reduced Hessian
2. `HessianLagrangian` is storing `Vector` in its attributes, whereas `BatchHessianLagrangian` is storing `Matrix`

src/Evaluators/common.jl
frapac: It makes sense to create a new kernel for that, indeed. In fact, we have already implemented that kernel in #181 : 
https://github.com/exanauts/ExaPF.jl/blob/develop/src/architectures.jl#L15

Project.toml
frapac: This is resolved

frapac:
Closing in favor of #185

Issue #80, opened 2020-12-21
Disallow scalar operations on GPU 


frapac:
I believe we could get some improvements by ensuring that no scalar operations are performed on the GPU.

frapac:
Solved by #181 and #186

PR #181, opened 2021-06-28
Disallow scalar indexing on GPU - part I

frapac:
With this PR, all the operations in `src/Polar/` do not use scalar indexing anymore. Now, the following code works: 
```julia
julia> CUDA.allowscalar(false)

julia> TestPolarFormulation.runtests("data/case9.m", CUDADevice(), CuArray)
Test Summary: | Pass  Total
PolarForm API |   50     50
Test Summary:      | Pass  Total
PolarForm AutoDiff |   46     46
Test Summary:      | Pass  Broken  Total
PolarForm Gradient |    5       1      6
Test Summary:      | Pass  Total
PolarForm Hessians |   13     13
Test Summary:    |
Batch algorithms | No tests
Test.DefaultTestSet("Batch algorithms", Any[], 0, false, false)
```

Before fixing the `Evaluators` code, I prefer to wait till #179 is merged, as this is a thorough refactoring of the Hessian code.

Ref #80

### DIFF
--- a/src/Evaluators/common.jl
+++ b/src/Evaluators/common.jl
@@ -77,6 +77,7 @@ function hessian(nlp::AbstractNLPEvaluator, x)
     return H
 end
 
+mynorm(a) = maximum(abs.(a))
 # Counters
 abstract type AbstractCounter end
 
@@ -102,8 +103,8 @@ function _check(val, val_min, val_max)
     violated_sup = findall(val .> val_max)
     n_inf = length(violated_inf)
     n_sup = length(violated_sup)
-    err_inf = norm(val_min[violated_inf] .- val[violated_inf], Inf)
-    err_sup = norm(val[violated_sup] .- val_max[violated_sup] , Inf)
+    err_inf = mynorm(val_min[violated_inf] .- val[violated_inf])
+    err_sup = mynorm(val[violated_sup] .- val_max[violated_sup])
     return (n_inf, err_inf, n_sup, err_sup)
 end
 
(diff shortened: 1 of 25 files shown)


src/Evaluators/common.jl
```diff
@@ -102,8 +102,8 @@ function _check(val, val_min, val_max)
     violated_sup = findall(val .> val_max)
     n_inf = length(violated_inf)
     n_sup = length(violated_sup)
-    err_inf = norm(val_min[violated_inf] .- val[violated_inf], Inf)
-    err_sup = norm(val[violated_sup] .- val_max[violated_sup] , Inf)
+    err_inf = xnorm_inf(val_min[violated_inf] .- val[violated_inf])
+    err_sup = xnorm_inf(val[violated_sup] .- val_max[violated_sup])
```
michel2323, line 106: This is really slow on the GPU. I guess `nrm2` won't work? If so, maybe one can do an optional argument.

src/Polar/Constraints/active_power.jl
```diff
@@ -117,15 +83,13 @@ function matpower_jacobian(polar::PolarForm, X::Union{State,Control}, ::typeof(a
...
-    ref = polar.indexing.index_ref
+    ref, pv, pq = index_buses_host(polar)
     # Check consistency: currently only support a single slack node
     @assert length(λ) == 1
-    V = buffer.vmag .* exp.(im .* buffer.vang) |> Array
+    V = voltage_host(buffer)
     hxx, hxu, huu = PS.active_power_hessian(V, polar.network.Ybus, pv, pq, ref)
 
-    λₚ = λ[1]
+    λₚ = sum(λ)  # TODO
```
michel2323, line 92: `reduce(+, lambda)` I think works on the GPU.

michel2323 (approved):

src/Evaluators/common.jl
frapac: I agree. But as we are computing that only once, at the beginning, I think it won't be a bottleneck to use norm Inf here.

PR #187, opened 2021-07-08
Update to KA 0.7 and CUDA.jl 3.3

michel2323:
Two things:

* `] test` passes all tests, but `julia --project test/runtests.jl` fails on moonshot due to a kernel crashing. @frapac can you confirm this on your side? If so, we might have to investigate that.
* We used the trick `x[i:i]` to avoid scalar access warnings. Shouldn't we use `allowscalar` to explicitly allow it for a statement? That might be easier to track in the future.
* This also adds the `BlockPowerFlow` package. That was actually by accident, but now I'm wondering whether it should not be in `Project.toml`? Let me know your opinion and I'll rebase.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -20,11 +20,11 @@ SparseDiffTools = "47a9eef4-7e08-11e9-0b38-333d64bd3804"
 TimerOutputs = "a759f4b9-e2f1-59dc-863e-4aeb61b1ea8f"
 
 [compat]
-CUDA = "~2.6, ~3.2"
+CUDA = "~2.6, ~3.2, ~3.3"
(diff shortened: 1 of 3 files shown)


frapac (approved):
Looks good to me! 
- I don't have any crash when running the tests on moonshot. 
- I am all in to use `allowscalar` instead our little trick `x[i:i]`!
- I don't have any opinion whether ExaPF should depend on BlockPowerFlow or not. We should just be careful, as BlockPowerFlow is not a registered package...

michel2323:
Nice. I see you used the trick only twice, so I left it as is. And I removed the `BlockPowerFlow` dependency.

PR #190, opened 2021-07-22
Add script to benchmark power flow solver

frapac:
Add a script to benchmark the power flow solver directly from the terminal. 
Run simply as
```shell
julia --project benchmark/powerflow.jl
```
and gives a nice output, as 
```shell
------------------------------------------
case                 |      CPU |     CUDA
------------------------------------------
case9.m              |  0.00019 |  0.00384
case30.m             |  0.00033 |  0.00379
case57.m             |  0.00074 |  0.00510
case300.m            |  0.00487 |  0.00971
case1354.m           |  0.02372 |  0.01745
case9241pegase.m     |  0.44299 |  0.10131
------------------------------------------
```

Currently only direct linear solvers are supported.

### DIFF
--- a/benchmark/powerflow.jl
+++ b/benchmark/powerflow.jl
@@ -1,0 +1,96 @@
+using ExaPF
+using Printf
+using SparseArrays
+using TimerOutputs
+import ExaPF: LinearSolvers
+const LS = LinearSolvers
+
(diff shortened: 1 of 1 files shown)


michel2323 (approved):
Nice!

PR #201, opened 2021-09-03
Another look at the line flow gradient

michel2323:
* The tests of line flow gradient fail with the default tolerance of `rtol=1e-8`. I made both tolerances as tight as possible.
* Fixed a minor bug where arrays were uninitialized. This triggered the test too.
* I also adapted the kernel to just change the function `cos` to some other function both for AD and handwritten. The gradient test then passes.

### DIFF
--- a/src/Polar/kernels.jl
+++ b/src/Polar/kernels.jl
@@ -649,6 +649,8 @@ KA.@kernel function adj_branch_flow_edge_kernel!(
     Δθ = vang[fr_bus, j] - vang[to_bus, j]
     cosθ = cos(Δθ)
     sinθ = sin(Δθ)
+    dcosθ = -sinθ
+    dsinθ = cosθ
 
     # branch apparent power limits - from bus
     yff_abs = yff_re[ℓ]^2 + yff_im[ℓ]^2
@@ -698,8 +700,8 @@ KA.@kernel function adj_branch_flow_edge_kernel!(
     adj_cosθ += 2.0 * vmag[to_bus, j] * vmag[fr_bus, j]^3 *   yre_fr  * adj_from_flow
     adj_sinθ += 2.0 * vmag[to_bus, j] * vmag[fr_bus, j]^3 * (-yim_fr) * adj_from_flow
 
-    adj_Δθ =   cosθ * adj_sinθ
-    adj_Δθ -=  sinθ * adj_cosθ
+    adj_Δθ =   dsinθ * adj_sinθ
+    adj_Δθ +=  dcosθ * adj_cosθ
     adj_va_from_lines[ℓ, j] += adj_Δθ
     adj_va_to_lines[ℓ, j] -= adj_Δθ
 end
--- a/test/Polar/gradient.jl
+++ b/test/Polar/gradient.jl
@@ -85,16 +85,20 @@ function test_line_flow_gradient(polar, device, MT)
         ExaPF.flow_constraints(polar, g, adcache)
         return sum(g)
     end
-    # adgradg = ForwardDiff.gradient(sum_constraints,x)
+    adgradg = []
+    CUDA.@allowscalar adgradg = ForwardDiff.gradient(sum_constraints,x)
     fdgradg = FiniteDiff.finite_difference_jacobian(sum_constraints,x)
     ## We pick sum() as the reduction function. This could be a mask function for active set or some log(x) for lumping.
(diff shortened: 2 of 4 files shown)


src/Polar/kernels.jl
```diff
@@ -649,6 +649,8 @@ KA.@kernel function adj_branch_flow_edge_kernel!(
     Δθ = vang[fr_bus, j] - vang[to_bus, j]
     cosθ = cos(Δθ)
     sinθ = sin(Δθ)
+    dcosθ = -sin(Δθ)
```
frapac, line 652: I am not sure about that, as `sin` is an expensive operation. How about doing instead?
```julia
    cosθ = cos(Δθ)
    sinθ = sin(Δθ)
    dcosθ = -sinθ
    dsinθ = cosθ
```

test/Polar/gradient.jl
```diff
@@ -85,16 +85,18 @@ function test_line_flow_gradient(polar, device, MT)
         ExaPF.flow_constraints(polar, g, adcache)
         return sum(g)
     end
-    # adgradg = ForwardDiff.gradient(sum_constraints,x)
+    adgradg = ForwardDiff.gradient(sum_constraints,x)
```
frapac, line 88: Right, we should uncomment this line. This line was commented previously as evaluating `ForwardDiff.gradient` on `sum_constraints` breads with `CUDA.@allowscalar false`

test/Polar/gradient.jl
```diff
@@ -85,16 +85,18 @@ function test_line_flow_gradient(polar, device, MT)
...
     fdgradg = FiniteDiff.finite_difference_jacobian(sum_constraints,x)
     ## We pick sum() as the reduction function. This could be a mask function for active set or some log(x) for lumping.
     m_flows = ExaPF.size_constraint(polar, ExaPF.flow_constraints)
     weights = ones(m_flows) |> MT
     gradg = similar(u, 2 * nbus)
     ExaPF.flow_constraints_grad!(polar, gradg, cache, weights)
-    # @test isapprox(adgradg, fdgradg)
-    # TODO: The gradient is slightly off with the handwritten adjoint
-    @test_broken isapprox(gradg, fdgradg[:])
+    # Both tests fail with rtoldefault(...) = 1e-8
```
frapac, line 95: I think `1e-8` is too low a tolerance

test/Polar/gradient.jl
```diff
@@ -85,16 +85,18 @@ function test_line_flow_gradient(polar, device, MT)
...
+    # The handwritten adjoint disagrees with AutoDiff!
+    @show isapprox.(adgradg, gradg, rtol=1e-6)
```
frapac, line 97: but at least this test is passing right? Except that the tolerance is set equal to `1e-6`?

test/Polar/gradient.jl
```diff
@@ -85,16 +85,18 @@ function test_line_flow_gradient(polar, device, MT)
...
+    # The finite difference gradient disagrees with AutoDiff!
+    @show isapprox.(gradg, fdgradg[:], rtol=1e-5)
```
frapac, line 99: That, that, I don't understand

frapac (commented):
A first stone to kill the bird!

src/Polar/kernels.jl
michel2323: Solely for debugging. This is not ready to merge in fact. Sorry for requesting review.

test/Polar/gradient.jl
michel2323: Yes.

test/Polar/gradient.jl
michel2323: Ah should be `@test all(...)`. I want to test the entries not the norm of the vectors.

michel2323:
@frapac I did another check, term by term. I am pretty sure there is nothing wrong with the adjoint.

test/Polar/gradient.jl
michel2323: That's the default for `approx()`.

test/Polar/gradient.jl
michel2323: Yes, so I allowed it now. Or should we disable the test with CUDA. In general, it's good to at least test here with AD on the CPU.

src/Polar/kernels.jl
michel2323: Fixed.

frapac (approved):

PR #202, opened 2021-09-27
Fix breaking Krylov@0.7.7 interface change

michel2323:


src/LinearSolvers/LinearSolvers.jl
```diff
@@ -251,9 +251,10 @@ function ldiv!(solver::DQGMRES,
 )
     P = solver.precond.P
     CUDA.allowscalar() do
-        global (y[:], status) = Krylov.dqgmres!(solver.inner, J, x; N=P)
+        Krylov.dqgmres!(solver.inner, J, x; N=P)
```
frapac, line 254: ouch ... why did we have a `global` here?

frapac (approved):
Looks good to me. One (not directly related) question: why do we need `allowscalar` to use `dqgmres` and `bicgstab` on GPU? I am surprised that not all operations are vectorized?

src/LinearSolvers/LinearSolvers.jl
michel2323: Because of the do loop with its own namespace. I don't think `global` is a problem there or? I thought it only moves the variable one scope up in this case.

frapac: In my understanding it becomes a global variable available everywhere ^^ 
In that case, I prefer to use the keyword `local`

PR #203, opened 2021-10-04
[exaopt] Integrate properly line flow constraints in optimization routine

frapac:
- [breaking] Update interface for `DirectSolver`. Now the constructor requires a sparse matrix `J` to be instantiated: `linsol = DirectSolver(J)` (explicit is better than implicit). 
- Add efficient evaluation of line flow's Hessian (cf new function `network_line_operations`). 
- fix line flow's adjoint in batch mode (adjoint is 1D, not 2D)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -27,7 +27,7 @@ LightGraphs = "1.3"
 Metis = "1"
 SparseDiffTools = "1"
 TimerOutputs = "0.5"
-julia = "1.7"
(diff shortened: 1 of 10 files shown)


frapac:
Tests are breaking because we need `cusolverRF` on the GPU. See https://github.com/JuliaGPU/CUDA.jl/pull/828

frapac:
Close stalling PR.

PR #211, opened 2021-11-29
hotfix: Removing KA for CPU Block Jacobi preconditioner

michel2323:
This should fix the random segfaults and linear solver crashes.

### DIFF
--- a/src/LinearSolvers/preconditioners.jl
+++ b/src/LinearSolvers/preconditioners.jl
@@ -260,37 +260,6 @@ function update(p, J::Transpose{T, CUSPARSE.CuSparseMatrixCSR{T}}, device) where
     _update_gpu(p, Jt.colPtr, Jt.rowVal, Jt.nzVal, device)
 end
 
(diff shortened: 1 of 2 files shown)


src/LinearSolvers/preconditioners.jl
```diff
@@ -300,10 +269,35 @@ Note that this implements the same algorithm as for the GPU and becomes very slo
 
 """
 function update(p, J::SparseMatrixCSC, device)
-    kernel! = update_cpu_kernel!(device)
     p.P.nzval .= 0.0
-    ev = kernel!(J.colptr, J.rowval, J.nzval, p, p.lpartitions, ndrange=p.nblocks, dependencies=Event(device))
-    wait(ev)
+    # Fill Block Jacobi
+    # TODO: Enabling threading leads to a crash here
+    for b in 1:p.nblocks
```
frapac, line 275: Looks good to me. I think this code can be optimized further on the CPU, but that's not the scope of this PR. Maybe should we open an issue about this?

frapac (approved):
Nice catch!

PR #213, opened 2021-12-23
Add nonlinear basis for vectorized OPF formulation 

frapac:
- Add the nonlinear basis introduced in https://arxiv.org/pdf/1906.09483
- Rework the adjoint accumulation for the line flow constraints


frapac:
Superseded by #214

Issue #199, opened 2021-08-31
Fix evaluation of `voltage_magnitude_constraints` on GPU


frapac:
The following code leads to inefficient execution on the GPU: 
https://github.com/exanauts/ExaPF.jl/blob/master/src/Polar/Constraints/voltage_magnitude.jl#L8

We should implement a kernel to do run that operation smoothly on GPU.

frapac:
Resolved in new formulation #214

Issue #85, opened 2021-01-05
Allow for flexible indexing of buses


dmaldona:
The Jacobian matrix structure depends on the indexing of the buses and the way we order the variable vector (i.e. [v1 v2 a1 a2] or [v1 a1 v2 a2], etc.). Having control over this might help us both enhance the performance of the linear solver and decrease divergent scenarios.

Reviewers: @michel2323 , @frapac

PR #214, opened 2022-01-12
Refactoring of ExaPF

frapac:
- simplify drastically the code (would help future maintenance)
- all functions are now implemented in `src/Polar/functions.jl` with the nonlinear basis introduced in https://arxiv.org/abs/1906.10652. This allows to streamline the evaluations of the callbacks with sparse linear algebra (mostly SpMV operations). 
- remove old code

A tutorial explaining the new implementation is coming.


frapac:
This PR addresses #199 #85 #28

PR #222, opened 2022-02-04
clean names in API

frapac:
- `VoltageMagnitudePQ` has been renamed `VoltageMagnitudeBounds`
- `my_map` has been renamed `mapping`
- add constructor `Jacobian(polar, func, State())`
- add underscore before names of private functions
- remove unused code
- clean tests & improve coverage


src/LinearSolvers/LinearSolvers.jl
```diff
@@ -160,7 +160,7 @@ function ldiv!(solver::BICGSTAB,
     y[:], n_iters, status = bicgstab(J, x, solver.precond, y; maxiter=solver.maxiter,
                                      verbose=solver.verbose, tol=solver.tol)
     if status != Converged
-        @warn("BICGSTAB failed to converge. Final status is $(status)")
+        error("BICGSTAB failed to converge. Final status is $(status)")
```
michel2323, line 163: `error`??? You are rough :-) . Maybe one wants to continue doing something no?

frapac: Right ... What do you think it's better? Issuing a warning with `@warn` or log something with `println`?

michel2323: I found `@warn` appropriate. I'd say `@info` or `@warn`. I think I'd prefer `@warn` because one may want to turn off normal output on larger systems, but may still want to be warned about this.

frapac:
Lol now the tests are breaking if we don't update the doc. But I think this is appropriate

michel2323:
> Lol now the tests are breaking if we don't update the doc. But I think this is appropriate

Oh yes!

PR #224, opened 2022-02-08
hotfix: replace my_map by mapping

frapac:
Don't know why this was not caught in the tests before...


michel2323:
I was thinking... weird name for Francois :-) .

PR #226, opened 2022-02-09
Refactor benchmark suite in ExaPF

frapac:
Major overhaul of the benchmark folder. Now, we benchmark: 
- the evaluation of the callbacks (functions, Jacobian, Hessian)
- the power flow algorithm 
- the BICGSTAB algorithm with the overlapping Schwarz preconditioner 

To launch the benchmark suite in the ExaPF folder: 
```shell
make benchmark
```

The default config is stored in `benchmark/config.json`. By default we benchmark ExaPF on `case1354pegase` and `case9241pegase`. The results are dumped as CSV files in `benchmark/results/`.


test/runtests.jl
```diff
@@ -67,14 +67,6 @@ init_time = time()
...
 
     include("quickstart.jl")
-
-    @testset "Test Benchmark script" begin
-        empty!(ARGS)
-        push!(ARGS, "KrylovBICGSTAB")
-        push!(ARGS, "CPU")
-        push!(ARGS, joinpath(INSTANCES_DIR, "case300.m"))
-        include(joinpath(BENCHMARK_DIR, "benchmarks.jl"))
-    end
```
michel2323, line 77: Shouldn't we test the benchmark code? Just with a small example...

michel2323 (approved):

test/runtests.jl
frapac: no you are right, we should test the benchmark code indeed

dmaldona (approved):

PR #227, opened 2022-02-10
Improve tests coverage 

frapac:
- test each KA's kernel on the CPU and verify output
- test preconditioner with overlap > 0


frapac:
Codecov is not working anymore, apparently

michel2323 (approved):
🧹

PR #240, opened 2022-04-08
Streamline evaluation of derivatives in block 

frapac:
- add a new stack `BlockNetworkStack` extending `NetworkStack` to `nblocks` different set of parameters (=loads)
- add `BatchJacobian` and `BatchHessian` to evaluate sparse Jacobians and Hessians in batches
- add a function `blockmul!` to streamline `mul!` if we are multiplying matrices inside of vectors (name is up for grab)


src/Polar/polar.jl
```diff
@@ -57,6 +57,42 @@ function mapping(polar::PolarForm, ::Control)
...
+    pf = polar.network
+    nbus = get(polar, PS.NumberOfBuses())
+    ref, pv, pq = pf.ref, pf.pv, pf.pq
+    nu = (length(pv) + 2*length(pq)) * k
+    # TODO
+    mapu = zeros(Int, nu)
+
+    shift_ang = k * nbus
+    shift_mag = 0
+    index = 1
```
michel2323, line 93: Why does this not return `mapu`?

michel2323 (approved):

PR #244, opened 2022-05-09
fix missing Event in KA kernels

frapac:
add `Event(device)` every time we are calling a kernel in ExaPF.


michel2323 (approved):
Oh sorry. I should have seen this in the review.

PR #248, opened 2022-05-16
[hotfix] Fix unitialized tape in NetworkStack

frapac:
Explain why the tests were failing randomly on certain CI runs...


michel2323 (commented):
@frapac Maybe we can do `VD(zeros(k*nlines))`, or `zeros(VD, k*nlines)`. Not sure the latter works. But I start to not like this `undef` in the code as a matter of principle.

PR #249, opened 2022-05-19
Get rid of uninitialized Arrays

michel2323:
Removed `undef` everywhere. Uninitialized variables can lead to bugs in the derivatives that are very hard to debug. I would be in favor or not using `undef` as a matter of style.

### DIFF
--- a/src/LinearSolvers/preconditioners.jl
+++ b/src/LinearSolvers/preconditioners.jl
@@ -119,9 +119,9 @@ struct BlockJacobiPreconditioner{AT,GAT,VI,GVI,MT,GMT,MI,GMI,SMT,VF,GVF} <: Abst
         end
         lpartitions = length.(partitions)
         blocksize = maximum(length.(partitions))
-        blocks = AT(undef, blocksize, blocksize, npart)
+        blocks = AT(zeros(eltype(AT), blocksize, blocksize, npart))
         # Get partitions into bit typed structure
-        bpartitions = MI(undef, blocksize, npart)
+        bpartitions = MI(zeros(eltype(MI), blocksize, npart))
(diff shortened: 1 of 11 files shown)


frapac (changes requested):
Right now, I am against merging this PR as is, as:
1. we introduce a lot of transfer between the host and the device (in fact, each time we instantiate an array)
2. the code is difficult to read, with a lot of duplicates. 

I would prefer we define a custom constructor for arrays, and then use this constructor consistently. To me, the constructor would be: 
```julia
function exalloc(VD::Type, dims...)
    arr = VD(undef, dims...)
    fill!(arr, zero(eltype(VD))
    return arr
end
```

michel2323:
@frapac With 
```julia
using CUDA
using BenchmarkTools
AT = CuArray{Float64}

function exalloc(VD::Type, dims...)
    arr = VD(undef, dims...)
    fill!(arr, zero(eltype(VD)))
    return arr
end

@btime cux1 = exalloc(AT, (10,))
@btime cux2 = CuArray(zeros(eltype(AT),10))
```
I get:
```
  5.161 μs (29 allocations: 1.23 KiB)
  6.669 μs (5 allocations: 304 bytes)
```
I think the GC and CUDA.jl are smarter than that, no? Maybe, we don't need the `fill!`.

This leads to the same number of allocations though:
```julia
@inline function exalloc(VD::Type, dims...)
    arr = VD(zeros(eltype(VD),dims...))
    return arr
end
```

However, I admit, I'm not a fan of adding another layer just for that, even if it adds duplicates throughout the code.

michel2323:
Revisit some other time I'd say.

PR #257, opened 2022-06-27
allow to load MATPOWER instances directly from PGLIB/MATPOWER

frapac:
- add PGLIB benchmark in Artifacts.toml
- add function `PS.load_case` to load `PowerNetwork` from case's name
- add function `load_polar` to load `PolarForm` from case's name

### DIFF
--- a/src/PowerSystem/power_network.jl
+++ b/src/PowerSystem/power_network.jl
@@ -119,6 +119,46 @@ function PowerNetwork(datafile::String; options...)
     return PowerNetwork(data; options...)
 end
 
+# Wrap ExaData and PGLIB's artifacts
+@enum(PowerNetworkLibrary,
+    EXADATA=1,
+    PGLIB=2,
+)
+
+"""
+    load_case(case_name, lib::PowerNetworkLibrary=EXADATA)
+
+Convenient function to load a [`PowerNetwork`](@ref) instance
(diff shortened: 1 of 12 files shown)


src/PowerSystem/power_network.jl
```diff
@@ -119,6 +119,46 @@ function PowerNetwork(datafile::String; options...)
...
+"""
+function load_case(case_name, lib::PowerNetworkLibrary=EXADATA)
+    src = endswith(case_name, ".m") ? case_name : "$(case_name).m"
+    if lib == EXADATA
+        src_dir = joinpath(artifact"ExaData", "ExaData")
+    elseif lib == PGLIB
+        src_dir = joinpath(artifact"PGLib_opf", "pglib-opf-21.07")
+        src = startswith(src, "pglib_opf") ? src : "pglib_opf_$(src)"
+    end
+    return PowerNetwork(joinpath(src_dir, src))
```
michel2323, line 159: This I am not sure I understand. It adds `LazyArtifacts.jl` as a dependency whereas it should IMO only be used in the tests. Plus one should still be able to load any local file no?

michel2323 (commented):
My only concern is adding the artifact as a dependency. I think the data artifacts should only be used for the tests.

michel2323 (approved):

PR #271, opened 2023-04-19
Change master to main

michel2323:
This doesn't need to be merge. Renaming the master branch as main to be consistent with our other projects and the general ecosystem.


michel2323:
If it passes and you have no suggestions @frapac I'll point the default branch to main.

PR #272, opened 2023-05-02
Julia 1.9rc3 precompilation

michel2323:
Julia 1.9rc3 precompilation breaks with the given `__init__()` function. The CUDA wrappers are now always included.


frapac:
Looks good to me. It would make sense to include CUDA by default as this is our only backend right now. 

On a side note, I am not sure to understand why the tests are broken with Documenter. Currently investigating this.

PR #273, opened 2023-05-02
Bump CUDA.jl to 4.2

michel2323:
This branch is based on the `ms/julia1.9rc3` branch.

### DIFF
--- a/LocalPreferences.toml
+++ b/LocalPreferences.toml
@@ -1,2 +1,0 @@
-[CUDA_Runtime_jll]
-version = "11.8"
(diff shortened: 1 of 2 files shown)


frapac (commented):
Looks good to me. 
The tests are broken when generating the documentation. This does not seem related to CUDA.jl. It's the same bug we are currently encountering in #272 

As a side note, why dropping support for CUDA.jl 4.1 ?

michel2323:
> As a side note, why dropping support for CUDA.jl 4.1 ?

That's indeed a mistake. I always mess up the semantics of the versions in Project.toml.

PR #276, opened 2023-09-22
Support CUDA.jl 5.x

michel2323:


michel2323:
@JuliaRegistrator register

PR #279, opened 2023-10-23
CUDA extension for ExaPF

michel2323:


ext/ExaPFCUDAExt.jl
```diff
@@ -0,0 +1,30 @@
...
+using LinearAlgebra
+using KernelAbstractions
+using SparseArrays
+
+const KA = KernelAbstractions
+const LS = ExaPF.LinearSolvers
+const PS = ExaPF.PowerSystem
+const AD = ExaPF.AutoDiff
+
+ExaPF.LS.DirectSolver(J::CUSPARSE.CuSparseMatrixCSR; options...) = ExaPF.LS.DirectSolver(nothing)
```
frapac, line 22: maybe remove `ExaPF.` here, as `LS` is already defined on line 18? Same on the following lines

ext/cuda_wrapper.jl
```diff
@@ -31,34 +25,34 @@ function Base.unsafe_wrap(Atype::Type{CUDA.CuArray{T, 1, CUDA.Mem.DeviceBuffer}}
     unsafe_wrap(CUDA.CuArray{T, 1}, p, (dim,); own, ctx)
 end
 
-CuSparseMatrixCSR{Tv, Int32}(A::SparseMatrixCSC{Tv, Ti}) where {Tv, Ti} = CuSparseMatrixCSR(A)
+CUDA.CUSPARSE.CuSparseMatrixCSR{Tv, Int32}(A::SparseMatrixCSC{Tv, Ti}) where {Tv, Ti} = CuSparseMatrixCSR(A)
```
frapac, line 28: I don't think we need `CUDA` here, `CUSPARSE` should be sufficient right?

src/LinearSolvers/preconditioners.jl
```diff
@@ -233,30 +203,32 @@ function mul!(Y, C::BlockJacobiPreconditioner, B::Matrix{T}) where T
     end
 end
 
-function mul!(y, C::BlockJacobiPreconditioner, b::CuVector{T}) where T
+function mul!(y, C::BlockJacobiPreconditioner, b::AbstractVector{T}) where T
+    device = KA.get_backend(b)
```
frapac, line 207: nice, indeed we should do that now!

src/Polar/second_order.jl
```diff
@@ -404,13 +404,13 @@ function AutoDiff.partials!(hess::ArrowheadHessian)
     n = length(duals)
     duals_ = reshape(reinterpret(T, duals), N+1, n)
 
-    if isa(H, SparseMatrixCSC)
+    if !_iscsr(H)
```
frapac, line 407: I would prefer having two functions `iscsr` and `iscsc` and write the condition as 
```julia
if iscsc(H)
    baba
elseif iscsr(H)
    abab
else 
    error("Not defined")
end
```
what do you think?

frapac (approved):
I think this is a net improvement to ExaPF. Thank you for working on this!

src/Polar/second_order.jl
michel2323: Done.

